Issue #14, opened 2017-04-13
Setup Travis


khase:
Setup the Travis CI.

### Backend
- [X] Linter _(This is executed with the test-command)_
- [X] Tests

### Frontend
- [X] Linter
- [ ] Tests
- [X] End2End Tests
- [ ] use [Chrome 59+ headless](https://developers.google.com/web/updates/2017/04/headless-chrome) so we do not need xvfb
  - We have to check before that the testframeworks are compatible with Chrome 59+ (headless)

aeimer:
Basic Setup merged with #25

aeimer:
Frontend Tests need to be fixed... E2E Tests are Running.

aeimer:
Will close this issue
- chrome headless may be replaced by <https://saucelabs.com/open-source>
- the karma-angular-unit-tests have to be set up
- the codecoverage have to be setup

aeimer:
See #127 and #94

aeimer:
@thomassss can you help us to be able to run all tests and setup a codecov integration? I have tried it last semester, but I couldn't get it running...

PR #21, opened 2017-04-18
Add Travis CI

aeimer:
We have setup travis so far, the linting works for TypeScript and the api-tests are running.

Todos are:
- Fix ```npm test``` in ```app/webFrontend```
- Fix ```ng e2e``` in ```app/webFrontend```
- Add correct ulr for Gitter notfifications

See #14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 language: node_js
 node_js:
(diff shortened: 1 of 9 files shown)


aeimer:
We should use the ```squash and merge``` option for this MR...

PR #28, opened 2017-04-20
Fixing linting errors that prevent backend compilation

grosspersky:
Because linting errors now prevent backend compiliation these files had to be fixed.


lukaskorte:
Not necessary anymore, because of #30

PR #54, opened 2017-05-02
feature/31-license-crawler: Adds license crawler to builds on develop

aeimer:
# Request Title:
Add license crawler to travis builds on brach develop

## Description:
This add a nlf-crawl after the tests and print out all found apache-licenses.

## Improvements
We now know which modules use a apache license and can eg add them to the readme.

## Known Issues:
no


aeimer:
Can be merged, but is not essential. We could solve the "build only on develop and not on PR"-problem first?!
To run the script costs about 20 seconds...
This solves issue #34.

PR #65, opened 2017-05-08
refactor docker build an push

khase:
move docker build to .travis/docker.sh
push also docker images tagged with branch name
execute docker.sh only on branch master and develop if its not a pull request

trigger travis CI on master branch


aeimer:
It would be nice to have some `echo`s in the `docker.sh`-File, so it is easier to understand what going on...

PR #110, opened 2017-05-18
Improve user interface usability (answer layout horizontally)

bernd-hda:


aeimer:
@bernd-hda I think this PR is set up the wrong way!
Currently you want merge FROM `develop` INTO `feature/8-define-tasks`.
You probably want it the other way around...

Just switch the two branches of your PR.

PR #118, opened 2017-05-25
Feature/unit inheritance

grosspersky:
# Request Title:

## Description:
This pull request includes major changes in both back- and frontend. Merge with caution.

Major changes are:
Single unit controller to retrieve all types of units from on API path.
Additional controllers for specific units only needed for POST

The {*}UnitController classes replace the individual controller classes used before.

Creation of units now happens inline on /course/detail/{:id} instead of /course/edit/{:id}.

## Improvements
- Tasks are created as units
- Tasks are not created in backend anymore before saving
- Inline editing / creation of units
- Units have a 'progressable' attribute to mark them as counting towards the lecture / course progress
- Unit have a 'weight' attribute for sorting --> should be useful later when introducing drag and drop

## Known Issues:
From CRUD only CR is currently working.


grosspersky:
The TaskController is the controller used before rewriting Task to TaskUnit.

PR #143, opened 2017-06-20
Adds a access log constraint by useragent

aeimer:
# Request Title:

## Description:
This line of Nginx-Config should log to the access log only if the useragent of the client is allowed. This will prevent to fill the logs withcalls from eg uptimerobot. I have testet on the reverse-proxy and edited the currently running image of the geli-staging-frontend.

Don't worry about the log-path, it is the default path. It is symlinked to stdout:
```
root@d2a76559bd3b:/# ls -la /var/log/nginx/
total 8
drwxr-xr-x 1 root root 4096 May 30 17:09 .
drwxr-xr-x 1 root root 4096 May 30 17:09 ..
lrwxrwxrwx 1 root root   11 May 30 17:09 access.log -> /dev/stdout
lrwxrwxrwx 1 root root   11 May 30 17:09 error.log -> /dev/stderr
```

This PR resolves no specific issue.

### DIFF
--- a/.docker/web-frontend/Dockerfile
+++ b/.docker/web-frontend/Dockerfile
@@ -4,5 +4,6 @@ FROM nginx:1.13
 RUN rm /etc/nginx/conf.d/default.conf
 
 COPY .docker/web-frontend/project.conf /etc/nginx/conf.d/project.conf
+COPY .docker/web-frontend/map_ua.conf /etc/nginx/conf.d/map_ua.conf
 
 COPY app/webFrontend/dist /var/www/dist
--- a/.docker/web-frontend/map_ua.conf
+++ b/.docker/web-frontend/map_ua.conf
@@ -1,0 +1,8 @@
+# Decide if useragent should be logged or not
+map $http_user_agent $log_ua {
+  # This is a exmaple config to ignore the useragent from uptimerobot.com
+  # ~*UptimeRobot 0;
+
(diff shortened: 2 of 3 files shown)


aeimer:
We definitely shouldn't add this file with our config to the repo. We could add a "dummy" non restrictive map_ua.conf and overwrite this file with our own needs on the server. 

I can check what happens if the file does not exist...

aeimer:
Yep it will brake, I'll change this.

```
root@UVM_MPSE_eLearning:/opt# docker logs staging_web-frontend_1
172.19.0.5 - - [22/Jun/2017:09:21:44 +0200] "HEAD / HTTP/1.0" 200 0 "http://staging.geli.fbi.h-da.de" "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)" "-"
172.19.0.5 - - [22/Jun/2017:09:31:44 +0200] "HEAD / HTTP/1.0" 200 0 "http://staging.geli.fbi.h-da.de" "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)" "-"
2017/06/22 09:34:55 [emerg] 1#1: unknown "log_ua" variable
nginx: [emerg] unknown "log_ua" variable
```

aeimer:
So this fixed and we can merge

PR #144, opened 2017-06-21
Feature/41 add students based on a csv file

OliverNeff:
# Request Title:

## Description:
Implement functions for upload csv files in backend and parse whitelist users to course.


api/src/controllers/CourseController.ts
```diff
@@ -52,15 +69,40 @@ export class CourseController {
...
+          if (c.students.indexOf(user._id) < 0) {
+            c.students.push(user);
+          }
+          return c.save().then((course) => course.toObject());
+        });
+      });
+  }
+
+
+  @Post('/:id/whitelist')
```
grosspersky: Add ``@Authorized('teacher')`` here.

OliverNeff: We have a different schema because we have in our files maybe users which are not exist in our system. 
It should not be the same schema as an normal user because you can log in with it and create a massive security gap.

OliverNeff:
Merged devlop into branch. No merge conflicts seen.

api/src/controllers/CourseController.ts.orig
```diff
@@ -0,0 +1,119 @@
+import {Request} from 'express';
```
OliverNeff: done

api/package.json.orig
```diff
@@ -0,0 +1,89 @@
+{
```
OliverNeff: done

OliverNeff:
done

OliverNeff:
This was my question last week, is it ok for all operation systems? 
It should be work on windows. Can someone with mac or linux ios look after it?

OliverNeff:
Done, found some better version I hope as add 26 mimetypes for all Os.

PR #163, opened 2017-06-29
Fix current password input

khase:
changes "current password" input in user edit form to a password field.


grosspersky:
Is included in #198

Issue #178, opened 2017-07-30
Editing profile does not work as expected


d89:
When editing the profile as teacher without filling the fields "current password", "password" or "confirm password", I receive the following stack trace and a not-very userfriedly toast notifaction telling me about an http error 500.

`{"name":"ValidationError","message":"User validation failed: password: Path `password` is required.","stack":"MongooseError: User validation failed: \n    at ValidationError (/usr/src/app/node_modules/mongoose/lib/error/validation.js:27:11)\n    at model.Document.invalidate (/usr/src/app/node_modules/mongoose/lib/document.js:1592:32)\n    at /usr/src/app/node_modules/mongoose/lib/document.js:1464:17\n    at validate (/usr/src/app/node_modules/mongoose/lib/schematype.js:745:7)\n    at /usr/src/app/node_modules/mongoose/lib/schematype.js:790:11\n    at Array.forEach (native)\n    at SchemaString.SchemaType.doValidate (/usr/src/app/node_modules/mongoose/lib/schematype.js:750:19)\n    at /usr/src/app/node_modules/mongoose/lib/document.js:1462:9\n    at _combinedTickCallback (internal/process/next_tick.js:73:7)\n    at process._tickCallback (internal/process/next_tick.js:104:9)","errors":{"password":{"message":"Path `password` is required.","name":"ValidatorError","properties":{"type":"required","message":"Path `{PATH}` is required.","path":"password","value":""},"kind":"required","path":"password","value":""}},"_message":"User validation failed"}`

On top of that, "Current Password" is not validated when editing the password - I can put whatever I want, it is not validated.

grosspersky:
I'm not able to reproduce a toast message showing this stack trace. My toast only show's this message:
```
Cancel
Response with status: 500 Internal Server Error for URL: http://localhost:4200/api/users/59a475eb831ca1248cabc796
```
The backend server also doesn't show this message and only this error message:
```
MongooseError: User validation failed:
    at ValidationError (D:\Git\geli\api\node_modules\mongoose\lib\error\validation.js:27:11)
    at model.Document.invalidate (D:\Git\geli\api\node_modules\mongoose\lib\document.js:1592:32)
    at D:\Git\geli\api\node_modules\mongoose\lib\document.js:1464:17
    at validate (D:\Git\geli\api\node_modules\mongoose\lib\schematype.js:745:7)
    at D:\Git\geli\api\node_modules\mongoose\lib\schematype.js:790:11
    at Array.forEach (native)
[...]

Issue #173, opened 2017-06-30
Return 403 instead of 401


aeimer:
To tell the frontend that a user is not authenticated, we currently send a 401. We should send a 403 to redirect the user to the login page.

khase:
401 = Unauthorized
403 = Forbidden (Permission denied)

we should send a 403 if a User is trying to do something he isn't alowed to (edit a course he isn't part of the teacher team)

PR #198, opened 2017-09-15
Fix profile editing issues

grosspersky:
# Request Title:

## Description:
This PR fixes the issues reported in #178.

## Improvements
Regular user data can be updated without filling the password fields.
If password fields are filled the current password field now gets corretly validated.

## Known Issues:
None


khase:
Your PR also adresses changes mentioned in #173

PR #207, opened 2017-10-02
Feature/108 visual non production warning

### LINKED ISSUE #108: Add a visual difference for staging system
# User Story

### As a: 
Developer/Tester

### I want:
to distinguish if I'm on the live or staging-system

### so that:
I don't accidentically work on the live system or vice versa.

## Storypoints:
?

## Acceptance criteria:
- [x] It is directly and on every page visible, that I'm on the stagingsystem
- [x] The switch to turn on the 'stagemode' gets automatically turned on on the staging-system

## Additional info:
We could do this with:
- a different bg-color in the header
- a big text like "ATTENTION - YOU ARE ON STAGING"
- a Version Badge in header (right)

khase:
closes #108 
Introduces a new Environment variable -> NONPRODUCTIONWARNING
and a new API route "/" which shows "status: up" (to make sure it is working) and the nonproductionwarning if provided

The webFrontend will check this route and if a warning is provided shows it in the top center of the application and changes the color of the toolbar to red.


api/src/controllers/InfoController.ts
```diff
@@ -0,0 +1,11 @@
+import {Get, JsonController} from 'routing-controllers';
+import {Info} from '../models/Info';
+
+@JsonController('/')
```
grosspersky: We shouldn't use the base url for this check.
I think an api path like /enviroment better suites our needs.

api/src/models/Info.ts
```diff
@@ -0,0 +1,12 @@
+import {IInfo} from '../../../shared/models/IInfo';
```
grosspersky: Should be named 'Enviroment'

shared/models/IInfo.ts
```diff
@@ -0,0 +1,4 @@
+export interface IInfo {
```
grosspersky: Should be called IEnviroment.

api/src/config/main.ts
```diff
@@ -21,5 +21,7 @@ export default {
     user: process.env.MAILUSER || undefined,
     pass: process.env.MAILPASS || ''
   },
-  mailSender: process.env.MAILSENDER || 'no-reply@geli.edu'
+  mailSender: process.env.MAILSENDER || 'no-reply@geli.edu',
+
+  nonProductionWarning: process.env.NONPRODUCTIONWARNING || undefined
```
grosspersky: For avoiding possible conflicts with other modules we should prefix this variable with GELI. Also we should call this after the contents of this variable. Maybe a better name would be something short like: GELI_ENVIROMENT.
Possible values would be:
- develop
- staging
- production
- demo

api/src/controllers/InfoController.ts
khase: This route wasn't intended to pass environmental variables to the frontend instead is should give some "static" informations about the API. I got inspired by the elasticSearch API which returns at its base URL something like this:
```
{
  "name" : "Leap-Frog",
  "cluster_name" : "elasticsearch",
  "version" : {
    "number" : "2.4.0",
    "build_hash" : "ce9f0c7394dee074091dd1bc4e9469251181fc55",
    "build_timestamp" : "2016-08-29T09:14:17Z",
    "build_snapshot" : false,
    "lucene_version" : "5.5.2"
  },
  "tagline" : "You Know, for Search"
}
```

It would be nice if we add here at least the version number of the build and maybe a timestamp and the git commit hash of the build.

If we dont want to use the base URL (which in my oppinion makes definitly sense for this application) we could move this to /about which isnt currently used.

api/src/config/main.ts
khase: This variable is intended to hold a fully customizable string which frontends or apps can react on like i did with our frontend coloring the head navbar read and displaying this message

aeimer:
We could add a ENV_VAR "LOGIN_INFO_BOX" which (if set) prints the given text as a box on the loginsite. This way every "customer" can say hello in their very own way AND we could print the credentials on the loginpage of our demo-site.

grosspersky:
I think a ENV_VAR is still not very userfriendly. After all this should be configurable through the admin interface, regarding text, color etc.

aeimer:
@grosspersky I'm definitely with you. Good idea!

khase:
@aeimer @grosspersky This is a great idea but should be an extra issue. This "feature" was ment to giva the chance of defining a pre defined "statemessage" for the API not giving the possibility to pass environmental Variables to the Frontend or Display any information on the Login page. 

`
As a:

Developer/Tester

I want:

to distinguish if I'm on the live or staging-system

so that:

I don't accidentically work on the live system or vice versa.
`

aeimer:
I'm with you. @khase do you open a new ticket?

Issue #168, opened 2017-06-29
Tagged Builds do not push a image to docker


aeimer:
Happened here: https://travis-ci.org/h-da/geli/builds/248483192#L908
The docker build should build the app and push a new Tag to the dockerhub, instead the build and push of the image was skipped:
```bash
$ .travis/docker.sh
+++ Run docker build and publish. +++
+ WARNING: branch v0.2.0 is not whitelisted -> skipping docker build and publish
```

khase:
I pushed this tag manually.
https://hub.docker.com/r/hdafbi/geli-api/tags/
https://hub.docker.com/r/hdafbi/geli-web-frontend/tags/

I don't know how bash and travis behaves. 
I thought `if [ -z ${TRAVIS_TAG+x}  ]; then` should do the trick...
Maybe travis does not set the variable right or the bash line does not nork as we suspected. 

We should test this on my personal repo khase/geli

PR #208, opened 2017-10-02
fixes issue #168 Tagged Builds do not push a image to docker

khase:
If the build isn't triggered trough a TAG the env TRAVIS_TAG is empty or not set.
So if the TRAVIS_TAG matches the TRAVIS_BRANCH this indicates a tagged build which should end in a tagged image on the docker hub

### DIFF
--- a/.travis/docker.sh
+++ b/.travis/docker.sh
@@ -25,7 +25,7 @@ if [ "$TRAVIS_BRANCH" == "master" ] || [ "$TRAVIS_BRANCH" == "develop" ]; then
     echo -e "${YELLOW}+ WARNING: pull request #$TRAVIS_PULL_REQUEST -> skipping docker build and publish${NC}";
   fi
 else
-  if [ "$TRAVIS_BRANCH" == "$TRAVIS_TAG" ]; then
+  if [ -n "$TRAVIS_TAG" ]; then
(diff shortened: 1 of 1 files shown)


.travis/docker.sh
```diff
@@ -25,7 +25,7 @@ if [ "$TRAVIS_BRANCH" == "master" ] || [ "$TRAVIS_BRANCH" == "develop" ]; then
     echo -e "${YELLOW}+ WARNING: pull request #$TRAVIS_PULL_REQUEST -> skipping docker build and publish${NC}";
   fi
 else
-  if [ -z ${TRAVIS_TAG+x}  ]; then
+  if [ "$TRAVIS_BRANCH" == "$TRAVIS_TAG" ]; then
```
aeimer: I would suggest using 
`[ -n STRING ] or [ STRING ] => True if the length of "STRING" is non-zero.`
Suggested here https://stackoverflow.com/questions/27896816/how-to-check-if-master-branch-pushed-with-a-tag-under-travis
```
if [ -n "$TRAVIS_TAG" ]; then
```

khase: The travis documentation says for the TRAVIS_BRANCH env 
`for builds triggered by a tag, this is the same as the name of the tag (TRAVIS_TAG).`

But your solution would definitly be the more elegant way :+1:

PR #220, opened 2017-10-15
Fix coverage issues

grosspersky:
# Request Title:

## Description:
This PR solves problems with test coverage not being correctly recorded.
The problem was caused by gulp-mocha being incompatible with istanbul.

Calling istanbul natively and adding configurations files for both istanbul and mocha solves the problem.

## Improvements
Coverage is finally working

## Known Issues:
None


.travis/coveralls.sh
```diff
@@ -2,7 +2,8 @@
 
 # Variables
 MODULE_PATH=".travis/node_modules"
-BIN_PATH="coveralls/bin"
+# BIN_PATH="coveralls/bin"
+BIN_PATH=".bin"
```
aeimer: Which bin folder do you reference here?

.travis/coveralls.sh
```diff
@@ -29,7 +30,8 @@ fi
 cd ..
 
 echo "+ sending lcov file to coveralls"
-cat api/coverage/lcov.info | $MODULE_PATH/$BIN_PATH/coveralls.js -v
+sed -i 's/bin\///g' api/coverage/lcov.info && cat api/coverage/lcov.info | $MODULE_PATH/$BIN_PATH/coveralls
```
aeimer: What does this sed does?

.travis/coveralls.sh
aeimer: All right, git this one. Could you adapt the same thing here https://github.com/h-da/geli/blob/bugfix/coverage/.travis/license-crawler.sh#L10 ? thx

.travis/coveralls.sh
grosspersky: Take a look at https://de.wikipedia.org/wiki/Sed_(Unix) ;-)

aeimer: Thanks for the detailed explanation...

.travis/coveralls.sh
grosspersky: Included the fix for the license-crawler.sh

PR #221, opened 2017-10-16
refactor restricting default value

### LINKED ISSUE #217: E-Mail service
Hi,
I have a problem with my Email service.
i have installed Geli-platform on my laptop und would like to use it locally.

the setting of my email service is:

export default {
  // Secret key for JWT signing and encryption
  secret: process.env.SECRET || 'notSoSecret234oi23o423ooqnafsnaaslfj',

  // BaseUrl for Webfrontend
  baseurl: process.env.BASEURL || 'http://localhost:4200',

  // Database connection information
  database: `mongodb://${process.env.DB_HOST || 'localhost'}:27017`,

  // Setting port for server
  port: process.env.PORT || 3030,

  // Email configuration
  // for provider see https://nodemailer.com/smtp/well-known/
  // Use either Provider or SMTPServer/Port
  mailProvider: process.env.MAILPROVIDER || 'debugmail.io',
  mailSMTPServer: process.env.MAILSMTPSERVER || undefined,
  mailSMTPPort: process.env.MAILSMTPPORT || 25,
  mailAuth: {
    user: process.env.MAILUSER || 'my email used by debugmail.io ',
    pass: process.env.MAILPASS || 'my password'
  },
  mailSender: process.env.MAILSENDER || 'no-reply@geli.edu',
  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '@h-da.de$',
};

i get this message

error

and in my terminal : 
Error
api_1           |     at new HttpError (/usr/src/api/node_modules/routing-controllers/http-error/HttpError.js:27:23)
api_1           |     at User_1.User.findOne.then (/usr/src/api/build/src/controllers/AuthController.js:34:23)
api_1           |     at process._tickCallback (internal/process/next_tick.js:109:7)

thx

khase:
refactors restricting `TEACHER_MAIL_REGEX` to `$` to avoid trouble mentioned in #217 (closes #217)

### DIFF
--- a/api/src/config/main.ts
+++ b/api/src/config/main.ts
@@ -22,5 +22,5 @@ export default {
     pass: process.env.MAILPASS || ''
   },
   mailSender: process.env.MAILSENDER || 'no-reply@geli.edu',
-  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '$',
+  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '^.+@.+$',
 };
--- a/api/src/config/main.ts
+++ b/api/src/config/main.ts
@@ -22,5 +22,5 @@ export default {
     pass: process.env.MAILPASS || ''
   },
   mailSender: process.env.MAILSENDER || 'no-reply@geli.edu',
-  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '@h-da.de$',
+  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '$',
 };
--- a/api/src/config/main.ts
+++ b/api/src/config/main.ts
@@ -22,5 +22,5 @@ export default {
     pass: process.env.MAILPASS || ''
   },
   mailSender: process.env.MAILSENDER || 'no-reply@geli.edu',
-  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '^.+@.+$',
+  teacherMailRegex: process.env.TEACHER_MAIL_REGEX || '^.+@.+\..+$',
 };


khase:
should we change this to `.*$` ? In my oppinion at least the `$` should be present

grosspersky:
I'm not able to test it right now. Maybe the h_da WLAN is blocking contact with the debug mail smtp server. Shouldn't we also keep the ``@``, to have at least a little bit of email validation?

aeimer:
```
var strings = ['test', 'test@local', 'test@stud.h-da.de', 'test@h-da.de', '@local'];

strings.forEach(function(elem) {
  console.log(elem);
  console.log(elem.match('^.+@.+$'));
});
```
Output:
```
test
null
test@local
[ 'test@local', index: 0, input: 'test@local' ]
test@stud.h-da.de
[ 'test@stud.h-da.de', index: 0, input: 'test@stud.h-da.de' ]
test@h-da.de
[ 'test@h-da.de', index: 0, input: 'test@h-da.de' ]
@local
null
```

So I would suggest `^.+@.+$`

khase:
@grosspersky i don't think it is neccessary to check for a "valid" email here. This feature -> `teacherMailRegex` is to deny random people to register as a teacher.
The check if a email is "valid" should be somewhere in the frontend an not be possible to modify,

@aeimer in which case does a "email"-domain does not contain DOT?

aeimer:
So in general we HAVE to check everything in backend. We should check it in the frontend too, but as a final check. A attacker could manipulate the sent HTTP-Request. 

An company internal eLearning Platform with local Mailadresses may be a case like: `admin@localhost`. Or am I wrong here?

PR #251, opened 2017-10-24
Feature/187 confirm member deletion

### LINKED ISSUE #187: Delete confirmation should be required, when deleting a student from a course
When removing a student from a course, I want to be prompted if I really want to do this.

grosspersky:
## Description:
If a student or a teacher is to be removed from a course the user is presented a confirm dialog.
Closes #187 

## Improvements
- No more accidental removal of students or teachers.

## Known Issues:
- None


app/webFrontend/src/app/shared/services/dialog.service.ts
```diff
@@ -20,9 +20,17 @@ export class DialogService {
...
   }
 
-  public delete(itemType: string, itemName): Observable<boolean> {
+  public confirmDelete(itemType: string, itemName): Observable<boolean> {
     return this.confirm('Delete ' + itemType,
-      'Are you sure you want to delete the ' + itemType + ' ' + itemName,
+      'Are you sure you want to confirmDelete the ' + itemType + ' ' + itemName,
       'Delete');
   }
+
```
aeimer: Why do we have two functions `confirmDelete` and `confirmRemove`? Because TS cannot overload functions?

app/webFrontend/src/app/shared/services/dialog.service.ts
```diff
@@ -20,9 +20,17 @@ export class DialogService {
     return dialogRef.afterClosed();
   }
 
-  public delete(itemType: string, itemName): Observable<boolean> {
+  public confirmDelete(itemType: string, itemName): Observable<boolean> {
     return this.confirm('Delete ' + itemType,
-      'Are you sure you want to delete the ' + itemType + ' ' + itemName,
+      'Are you sure you want to confirmDelete the ' + itemType + ' ' + itemName,
```
aeimer: I think `confirmDelete` is wrong in this sentnce...

app/webFrontend/src/app/shared/services/dialog.service.ts
aeimer: If you want to have different functions because of the `Want to remove ... FROM ...`, then we should call the functions
- `confirmDelete`
- `confirmDeleteFrom`

app/webFrontend/src/app/shared/services/dialog.service.ts
grosspersky: This was bad refactoring...

grosspersky:
Now everything is working. :-)

PR #252, opened 2017-10-24
Remove Matrikelnr confirmation in Register view #244

### LINKED ISSUE #244: No need to confirm the Matrikelnummer in the Register view
# User Story

### As a: 
user

### I want:
to have only one field for the _Matrikelnummer_ in the Register view

### so that:
I don't need to enter the same (visible) value twice.

## Acceptance criteria:

- [x] Only one field for the _Matrikelnummer_ in the Register view
- [x] No confirmation of the _Matrikelnummer_ anymore

## Additional info:
This was discussed (and also already greenlighted by Prof. Trapp and @d89) during our meeting last Thursday.

micpah:
## Description:
Removed field "Confirm Matrikelnummer" from Register view.
Closes #244 

## Improvements
The user does not have to enter the _Matrikelnummer_ twice during the registration.

## Known Issues:
There has not been any validation (i.e. check for matching) for the _Matrikelnummer_ anyway (like it is for password). You could fill out two different numbers and still got registered successfully.


app/webFrontend/src/app/auth/register/register.component.ts
```diff
@@ -54,7 +54,6 @@ export class RegisterComponent implements OnInit {
       }),
       email: ['', Validators.required],
       uid: [null, Validators.required],
-      confirmUid: [null, Validators.required],
```
aeimer: Haven't `uid` and `uidConfirm` never been checked if they are the same like `password`?

micpah: No, there has not been a validation for uid. I was also surprised. Just doublechecked it on demo: you can fill out two different numbers and still get registered successfully.

aeimer:
@micpah merge it if you want

Issue #188, opened 2017-07-30
Dragging and Dropping Lectures in the "Edit Course" section can lead to the deletion of all the lectures


d89:
When messing around with the drag and drop feature while rearranging lectures and material within lectures, from time to time, stuff gets deleted.

OliverNeff:
This is also a problem by managing member is course. 
I think it is a problem how we use dracula in our context.

aeimer:
@OliverNeff haven't you fixed this Issue?

aeimer:
is #131 related?

d89:
Did not occur any more :). Good job.

Issue #131, opened 2017-06-07
Users are hiding from Dom by drag and drop


OliverNeff:
Sometimes users are removed from DOM, when you drag one user and drop it on the other in "Members of Course" settings. This could have something to do with member and user collection and some if statements.

d89:
Did not happen to me 👍

PR #255, opened 2017-10-25
Bugfix/drag and drop

OliverNeff:
## Description:

- Optimize css so there is a same look for drag and not drag elements, with except of gravatar image. 
- Fix Issues where are the dragging element ist lost from dom.
- Add a send mail link to students and teachers.
-  Fix the placeholder text.
- Drag and drop is not triggered anymore by the hole line rather by the drag icon.

## Known Issues:
#131 & #188


app/webFrontend/src/app/course/course-edit/course-user-list/course-user-list.component.html
```diff
@@ -1,5 +1,5 @@
 <md-input-container class="fuzzySearch">
-  <input mdInput placeholder="Search and add teachers by name, Id, e-mail"
+  <input mdInput placeholder="Search and add {{dragulaBagId}} by name, Id, e-mail"
```
lukaskorte: Do not output ids as text

app/webFrontend/src/app/course/course-edit/course-user-list/course-user-list.component.html
```diff
@@ -34,29 +34,29 @@
 <div class="users" *ngIf="fuzzySearch.length > 0">
   <p><b>In course ({{usersInCourse.length}})</b></p>
```
lukaskorte: Substitute `<b>` with `<strong>`

app/webFrontend/src/app/course/course-edit/course-user-list/course-user-list.component.html
OliverNeff: I add a role field in my components but it could be better, if we have a shared role enum in our project.

PR #274, opened 2017-10-28
Use different databases for regular development and tests

### LINKED ISSUE #135: Use different databases for prod, dev and testing
# User Story

### As a: 

Developer

### I want:

to use different databases for production, development and testing purposes

### so that:

no data is accidentally dropped and i can test with confidence

## Acceptance criteria:

- [x] api tests use their own test database
- [x] regular development api servers use their own dev database
- [x] production deployments use their own prod database

## Additional info:

eg. possible via ENV vars

felixbrucker:
## Description:

This PR will use `geli` as default db name instead of `test`. Additionally the db name can be set via the environment variable `DB_NAME` if needed. API tests will use the `test` db to not conflict with user data.

## Improvements

see #135 

## Known Issues:

As previously the default db name was `test` one needs to migrate to the "new" `geli` name for production data.

Fixes #135


khase:
Do we really need the module `cross-env`? We used the env `NODE_ENV` before to determine if we are running Tests 
```
/**

* For testing mocha will start express itself

*/

if (process.env.NODE_ENV !== 'test') {

new Server().start();

}

```

felixbrucker:
You are correct, currently `NODE_ENV` is set to 'test' in each integration test file on the top: `process.env.NODE_ENV = 'test';`

I suggest we remove this in each file and just set it once via cross-env, what do you think?

khase:
Okay, I think this would definitly be the more reliable Method assuming someone creating a new test-file and forgetting to set this env.

felixbrucker:
Alright, i have removed these unneeded lines.

api/src/config/main.ts
```diff
@@ -1,3 +1,7 @@
+let dbName = process.env.DB_NAME || 'geli';
+if (process.env.NODE_ENV === 'test') {
+  dbName = 'test';
+}
```
khase: This should be a configuration Object.
Consider moving ``dbName`` inside the actual Object and the code to the server.ts      
```
if (process.env.NODE_ENV === 'test') {
   dbName = 'test';
 }
mongoose.connect(config.database);
´``

Comment (felixbrucker) [reply api/src/config/main.ts]: I do not think the server should need to modify the config, the config module should set the proper db url itself. Furthermore `dbName` should not be modified at a later stage, as it's only used to setup the db url.

**Comment by grosspersky** [comment]
Instead of using cross-env we just could modify the mocha.opts file zu require an additional script:
```
// mocha.opts
--require test/mocha.env.js
```

In the referenced file we would just do this:
```
// mocha.env.js
process.env.NODE_ENV = 'test';
```
This way all the test would run in the correct enviroment and the test command would be nice and clean.

**Comment by felixbrucker** [comment]
Good point, i have updated the code

File: api/src/config/main.ts (hunk 1 above)
Comment (khase) [reply api/src/config/main.ts]: using the latest changes we can simply set the ``dbName`` env in ``test/mocha.env.js`` to get rid of the `if` in `api/src/config/main.ts`

**Comment by aeimer** [comment]
We need to rename our DBs if we merge this: MAybe with this three-liner https://stackoverflow.com/questions/9201832/how-do-you-rename-a-mongodb-database

File: api/test/mocha.env.js [hunk 2]
```diff
@@ -1 +1 @@
-process.env.NODE_ENV = 'test';
```
khase: our tests neet the env `NODE_ENV` to be set:
``` typescript
/**
 * For testing mocha will start express itself
 */
if (process.env.NODE_ENV !== 'test') {
  new Server().start();
}
```

Issue #264, opened 2017-10-26
Refactor Course View


PatrickSkowronek:
## Acceptance criteria:
- [x] Remove enroll button on enrolled courses
- [x] Make whole title on course section clickable
- [x] Add a badge to show how many courses are available

grosspersky:
Solved with PR #285.

PR #275, opened 2017-10-29
Bugfix/262 refactor code katas

### LINKED ISSUE #262: Refactor Code-Katas
- [x] Add Example Template for creation to guide Teacher.
- [x] Improve styling

### Issues

- [x] Everytime I give focus to the editor, part of the editor is selected and the whole screen scrolls up. Pretty weird behavior. Full HD with newest chrome.
- [x] If I remove the whole "validate" function and the call to the validate function (as teacher, when I define the kata), my kata passes. It should be treated as "invalid".
- [x] When I hit "save" the kata is saved and a toast message appears which is telling me "saved". However, if I did not put a name for the Kata, it is actually not saved. I would expect an error-toast to be shown instead of "saved", when the name is missing.
- [x] I can't solve the kata in the frontend, when I'm a teacher. Would be nice if we could show a hint telling me that I can't solve this, because I'm the teacher of this course or so.

khase:
## Description:
closes #262 

## Improvements
continuous line numbers on unitComponent
fixes error logging for user written Code 
add styling for ace-editor (mostly padding)
Ace-Editor on unitFormComponent does expand if necessary (please validate)

## Known Issues:
none


grosspersky:
This is the PR #262 and not #264 .

PR #277, opened 2017-10-29
Adapts Mobile First Scheme => Breakpoints etc

aeimer:
## Description:
Adds some css-tweaks for a better design fundamental

## Improvements
- Now we can use the breakpoint mixin
- now we use the mobile first scheme
- we are 4k ready

## Known Issues:
- we need to check (and adapt) all custom component scss files


app/webFrontend/src/app/app.component.scss
```diff
@@ -24,8 +24,8 @@
   font-family: Pacifico, cursive;
   text-decoration: none;
   color: white;
-  font-size: 2.2rem;
-  margin-bottom: .2rem;
+  font-size: 28px;
```
aeimer: @lukaskorte Why do we need a px value here?

app/webFrontend/src/app/start/dashboard/dashboard.component.scss
```diff
@@ -4,17 +4,8 @@
...
-  margin-right: 20px;
-
-  .details-button {
-    float: right;
-  }
-}
-
 .add-course-fab {
-  position: absolute;
+  position: fixed;
```
aeimer: This positions the btn outside of the "page" in 4k view. It just looks crap: 
grafik

app/webFrontend/src/styles.scss
```diff
@@ -25,15 +25,6 @@ $app-theme: mat-light-theme($app-primary, $app-accent, $app-warn);
...
   height: 100%;
-  font-size: 10px;
-
-  @media only screen and (min-width: $bp_width--md) {
-    font-size: 12px;
-  }
-
-  @media only screen and (min-width: $bp_width--4k) {
-    font-size: 16px;
-  }
```
aeimer: If we have bigger resolutions, we need a bigger font-size so that all the (r)em styled elements scale correctly and the website is nice to handle. In general we have to set a base font-size, otherwise every browser will adapt it's own?!

PR #281, opened 2017-10-30
Feature/239 make it not so easy to create a teacher account

AndreasSchroll:
## Description:
changed registration default-form to student
Now you need to click on a small link to get to the teacher register


PatrickSkowronek:
The problem is solved with the simple path. For now I think it would work, expect:

- The text to the teacher form should be "Not a student? click here"
- the spacing between the last input field and the text for the teacher is incorrect
- The title should be in capital letters

app/webFrontend/src/app/auth/register/register.component.html
```diff
@@ -82,5 +79,7 @@
       </md-input-container>
       <button md-raised-button color="primary" type="submit" [disabled]="loading">Register</button>
     </form>
+    <div class="div-change-role" *ngIf="role=='student'">Not a student? click <a href="javascript:void(0)" (click)="changeRole('teacher')">here</a>.</div>
```
grosspersky: ``click`` starts a new sentence. So this should be ``Click`` instead.
Also you should add an additional icon to show this part more cleary. Should be this icon:
https://material.io/icons/#ic_arrow_forward

app/webFrontend/src/app/auth/register/register.component.html
```diff
@@ -82,5 +79,7 @@
...
+    <div class="div-change-role" *ngIf="role=='teacher'"><a href="javascript:void(0)" (click)="changeRole('student')">back</a> </div>
```
grosspersky: ``back`` marks the start of the "sentence", so this should be ``Back``.
Here you should also add an icon: https://material.io/icons/#ic_arrow_back

PR #283, opened 2017-10-30
add active flag for user view

OliverNeff:
## Description:
active flag was not implemented for students.


OliverNeff:
Implement some tests.

api/src/models/User.ts
```diff
@@ -26,7 +26,7 @@ const userSchema = new mongoose.Schema({
       lowercase: true,
       unique: true,
       required: true,
-      validate: [ isEmail, 'invalid email' ]
+      validate: [{ validator: (value: any) => isEmail(value), msg: 'Invalid email.' }]
```
OliverNeff: look at: https://stackoverflow.com/questions/42877930/implicit-async-custom-validators-custom-validators-that-take-2-arguments-are-d

PR #286, opened 2017-10-31
Feature/236 logo

OliverNeff:
## Description:
Add hda fbi logo to geli.
Url and link is just hard coded because of time problems,
in an other issue we should make it variable.

## Known Issues:
236


OliverNeff:
I hope it is ok that I link the whole top-bar to hda-fbi. I thought so it is possible to change logo and link without conflicts.

lukaskorte:
Will need complete rework, closing for now

PR #291, opened 2017-11-01
Feature/89 allow admins to reset users passwords

### LINKED ISSUE #89: Password reset
# User Story

### As a: 
User

### I want:
a way to reset my password if i can't remember

### so that:
i am able to use my account again

## Acceptance criteria:
- [x] Reset link in login page 
- [x] send mail to user with reset link
- [x] reset link referes to a page where the user can set a new password
- [x] invalidate reset link if the user login succesfully / after given time (24h ??)
- [x] allow admins to change passwords of all users

khase:
## Improvements

Allows admins to change every users password without knowing the original one (except for his own)
close #89


app/webFrontend/src/app/admin/user-admin/user-admin.component.ts
```diff
@@ -76,4 +76,26 @@ export class UserAdminComponent implements OnInit {
...
     });
   }
+
+  resetUserPassword(userIndex: number) {
+    this.dialogService
+      .confirmUpdate('property', 'password', this.allUsers[userIndex].email)
+      .subscribe(res => {
+        if (res) {
+          const user = this.allUsers[userIndex];
+          user.password = 'haha';
```
grosspersky: This seems to be not really a secure solution.
You should create a random password in the backend and also inform the user who's password was changed, that it was changed and tell him the new one.

Or tell the admin the new, so that he can tell the user his new password. But this way nobody knows which new password is set.

PR #335, opened 2017-11-09
Bugfix/315 password constraint for reset password should be the same as for registration

### LINKED ISSUE #315: password constraint for "reset password" should be the same as for "registration"
It is possible to give yourself a weaker password when using the "reset password" functionality (or editing your profile) as when registering.

Please take care that we apply the same kind of error messages aswell.

AndreasSchroll:
## Description:
Password pattern moved to password.ts
added pattern to reset-password and change-password

## Known Issues:

closes #315


AndreasSchroll:
@lukaskorte 
I had the same idea, but the password field is not required in every form. such as user-edit

PR #354, opened 2017-11-15
fix accesskey bug

khase:
## Description:
Fixes bug which prevents user to join any course protected wit an accesskey

## Known Issues:
none


app/webFrontend/src/app/course/course-container/course-container.component.ts
```diff
@@ -43,7 +43,7 @@ export class CourseContainerComponent implements OnInit {
   enrollCallback({courseId, accessKey}) {
     this.courseService.enrollStudent(courseId, {
       user: this.userService.user,
-      accessKey
+      accessKey: accessKey
```
lukaskorte: Unnecesary

PR #356, opened 2017-11-15
Hotfix!!! fix accesskey bug

khase:
## Description:
Fixes bug which prevents user to join any course protected wit an accesskey

!!! First validate #354 !!!

## Known Issues:
none


lukaskorte:
Why do we need the same code in two different branches and PRs?

grosspersky:
This one is going directly into master to hotfix the live system.

PR #362, opened 2017-11-17
Now using maxWidth for Dialog via config, removes css-hack

aeimer:
## Description:
Removes global css hack for matDialog

## Improvements
- Now using MaterialOptions not a dirty css hack

## Known Issues:
_NONE_

------


app/webFrontend/src/_global-hacks.scss
```diff
@@ -1,5 +1,2 @@
 // With a material update we can disable this hack
 // see app/webFrontend/src/app/unit/unit-form/free-text-unit-form/free-text-unit-form.component.ts:78
```
khase: Wouldn't it make sense to delete the whole file since it's empty now?

aeimer: Yeah we could. But I though we could keep the file structure  for the following semesters...

PR #363, opened 2017-11-19
bugfix/#310-hidePWFields

### LINKED ISSUE #310: The field for the current password should be hidden in the "edit profile" section as admin
When I change the profile of another user as admin, I do not want to see the "password" field in the frontend ... because I don't know the current password.

HPunktOchs:
## Description:
The Password fields shall be hidden for everybody but the User who is allowed to see its own Password.

Closes #310 

## Improvements
- The Field is now hidden for those who are not allowed to see it.

## Known Issues:

------


khase:
Directly write the Issuenumber behind the Hashtag to link it
-> #310

app/webFrontend/src/app/user/user-edit/user-edit.component.ts
```diff
@@ -52,7 +57,7 @@ export class UserEditComponent implements OnInit {
   getUserData() {
     this.userDataService.readSingleItem(this.id).then(
       (val: any) => {
-        this.user = val;
+         this.user = val;
```
khase: please remove this unneccessary change

app/webFrontend/src/app/user/user-edit/user-edit.component.ts
```diff
@@ -33,15 +33,20 @@ export class UserEditComponent implements OnInit {
...
-  userIsAdmin(): boolean {
-    return this.userService.isAdmin();
+  hidePwFields() {
+    if (this.user) {
+      if (this.userService.user._id === this.user._id) {
+        return true;
+      }
+    } else {
+      return false;
+    }
```
khase: remove the `else` and always return `false` if you get to tis point.

Now you are just returning nothing if you are not the user which this profile belongs to (which 
incidental is the same as `false`)

app/webFrontend/src/app/user/user-edit/user-edit.component.ts
```diff
@@ -33,15 +33,20 @@ export class UserEditComponent implements OnInit {
     this.generateForm();
   }
 
-  userIsAdmin(): boolean {
-    return this.userService.isAdmin();
+  hidePwFields() {
```
khase: rename this to `showPwFields` to avoid confusion since the PwFields actually are shown when this validates to `true`

khase:
closes #370

PR #394, opened 2017-11-28
Feature/260 download units

### LINKED ISSUE #260: Student: Download selected units
## Acceptance criteria:

- [x] select all/unselect all
- [x] select one or more lectures or reasonable units (files, videos, markdown) in the current course
- [x] download a zip-file with an ordered list of all selected units
show progress (zip file generation progress): Is not possible 
- [x] implement a hierarchy folder structure, which display the folder <number>. <lectureName>/<number>. <unitName>
- [x] add a delete logic
- [x] add a cache logic
- [x] remove special character and allow only character and numbers
- [x] implement a zip size limit
- [x] add a proper styling
- [x] show items that were too large for including in the zip file and offer a seperate download (non-zipped)

PatrickSkowronek:
## Description:
Implemented a download dialog for unit, lectures and complete courses

Closes #260 

## Improvements
- New Download Dialog for everthing

## Known Issues:
temp folder will not be cleared
------


api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,93 @@
...
+  @Post('/')
+  async addTask(@Body() data: IDownload) {
+
+    if ( data.units.length === 0) {
+      throw new NotFoundError();
+    }
+
+    const fileName = await crypto.pseudoRandomBytes(16).toString('hex');
+    const filepath = appRoot + '/temp/' + fileName + '.zip';
+    const output = fs.createWriteStream(filepath);
```
grosspersky: Catch if file path is actually available and throw an error, currently the backend just crashes without any sign for the frontend.

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  @Post('/')
+  async addTask(@Body() data: IDownload) {
+
+    if ( data.units.length === 0) {
+      throw new NotFoundError();
+    }
+
+    const fileName = await crypto.pseudoRandomBytes(16).toString('hex');
+    const filepath = appRoot + '/temp/' + fileName + '.zip';
```
grosspersky: Consider moving the file path to the config file since it is referenced at three different locations. This will make further updates much easier und more configurable.

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,93 @@
...
+          archive.append(taskUnit.description + '\n' + fileStream, {name: taskUnit.name + '.txt'});
+        }
+      } else {
+        throw new NotFoundError();
+      }
+    }
+
+    archive.finalize();
+
+    return 'api/temp/' + fileName + '.zip';
```
grosspersky: Consider moving the file path to the config file since it is referenced at three different locations. This will make further updates much easier und more configurable.

api/src/server.ts
```diff
@@ -51,6 +51,7 @@ export class Server {
     }
 
     // TODO: Needs authentication in the future
+    this.app.use('/api/temp', express.static('temp'));
```
grosspersky: Consider moving the file path to the config file since it is referenced at three different locations. This will make further updates much easier und more configurable.

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,93 @@
...
+      zlib: {level: 9} // Sets the compression level.
+    });
+
+    archive.on( 'error', function(err: Error ) {
+      throw err;
+    });
+
+    archive.pipe(output);
+
+    for (const unit of data.units) {
```
grosspersky: This whole for loop should be reworked. Attach your export methods directly to the matching mongoose schemas. This way you can avoid this if/else thing.

app/webFrontend/src/app/app.module.ts
```diff
@@ -11,6 +11,7 @@ import {
   APIInfoService,
   CodeKataUnitService,
   CourseService,
+  DownloadReq,
```
grosspersky: Bad naming, should be called something like CourseDownloadService.

app/webFrontend/src/app/shared/services/data.service.ts
```diff
@@ -403,6 +404,21 @@ export class AboutDataService extends DataService {
...
+  });
+}
+}
+
+@Injectable()
+export class DownloadReq extends DataService {
+  constructor(public backendService: BackendService) {
+    super('download/', backendService);
+  }
+  postDownloadReqForCourse(idl: IDownload) {
```
grosspersky: Why do you need so additional route for this. Just use the createItem method of the data service

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,93 @@
...
+import {FileUnit} from '../models/units/FileUnit';
+import {VideoUnit} from '../models/units/VideoUnit';
+import {IVideoUnit} from '../../../shared/models/units/IVideoUnit';
+import {Task} from '../models/Task';
+
+@JsonController('/download')
+@UseBefore(passportJwtMiddleware)
+export class DownloadController {
+
+  @Post('/')
```
grosspersky: Try setting the content type to zip to activate the regular download dialog.

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,243 @@
...
+          const fileUnit = <IFileUnit><any>localUnit;
+          fileUnit.files.forEach((file, index) => {
+            if (unit.files.indexOf(index) > -1) {
+              if ((file.size / 1024 ) > 51200) {
+                localTooLargeFiles.push(file.path);
+              }
+              localTotalSize += (file.size / 1024 );
+            }
+          });
+        } else if (localUnit.__t === 'video') {
```
grosspersky: This type does no longer exists after the migration scripts have been executed.
Also your filesize calculation would be redudant for the two types.

So you should either remove this else-Branch or at least refactor the file size calculation into resusable function.

api/src/controllers/DownloadController.ts
```diff
@@ -0,0 +1,243 @@
...
+        const localUnit = await
+          Unit.findOne({_id: unit.unitId});
+        if (localUnit.__t === 'file') {
+          const fileUnit = <IFileUnit><any>localUnit;
+          fileUnit.files.forEach((file, index) => {
+            if (unit.files.indexOf(index) > -1) {
+              data = data + file.name;
+            }
+          });
+        } else if (localUnit.__t === 'video') {
```
grosspersky: Previous comment applies here too.

PR #402, opened 2017-11-28
Feature/#18 delete course

### LINKED ISSUE #18: Delete course
# User Story

### As:
teacher or admin  

### I want to:
delete or deactivate old courses from list of courses. 

### As a result:
I can clean up old courses so there is a better overwiew on current

### Acceptance criteria:
teacher or admin can delete a course, it will show in another seaction like an trash or will never show again.

HPunktOchs:
## Description:
As a Teacher or an Admin i want to be able to delete a Course.

Closes #18 

## Improvements
-A Teacher/Admin is now able to delete a Course from the Edit Course/screen. To prehibit an accidential deletion of acourse, a Confirmation Dialog has been added.

## Known Issues:
_NONE_

------


api/src/controllers/CourseController.ts
```diff
@@ -211,12 +211,31 @@ export class CourseController {
...
+
+  @Authorized(['teacher', 'admin'])
+  @Delete('/:id')
+  async deleteCourse(@Param('id') id: string, @CurrentUser() currentUser: IUser) {
+    const conditions: any = {_id: id};
+    conditions.$or = [
+      {teachers: currentUser._id},
+      {courseAdmin: currentUser._id}
+    ];
+    const course = await Course.findOneAndRemove(conditions);
```
khase: Just deleting the Course won't do the trick you need to delete also all containing Lectures and Units recursively

api/src/controllers/CourseController.ts
```diff
@@ -211,12 +211,31 @@ export class CourseController {
...
+    if (course) {
+      return course;
```
khase: It doesn't really make any sense to return the course you've just deleted

api/test/integration/course.ts
```diff
@@ -202,4 +202,51 @@ describe('Course', () => {
...
+          description: 'Test description',
+          courseAdmin: user._id,
+          active: true
+        });
+      testData.save((error, course) => {
+        chai.request(app)
+          .del(`${BASE_URL}/${course._id}`)
+          .set('Authorization', `JWT ${JwtUtils.generateToken(user)}`)
+          .end((err, res) => {
+            res.status.should.be.equal(200);
```
khase: you should query for the Database to check if the deletion was successfull

api/src/controllers/CourseController.ts
grosspersky: This could call the defined pre-remove hook:
```
// Cascade delete
courseSchema.pre('remove', function (next: () => void) {
  Lecture.find({'_id': {$in: this.lectures}}).exec()
    .then((lectures) => Promise.all(lectures.map(lecture => lecture.remove())))
    .then(next)
    .catch(next);
});
```
But we should check if this is really happening or if we have to add this function to the pre 'findOneAndRemove' hook

Issue #50, opened 2017-05-01
deadlines within questions and assignments


utetrapp:
# User Story

### As a: 
teacher

### I want:
set a deadline for a given assignment/question

### so that:
able to address the problems in the lecture.

## Storypoints:
?

## Acceptance criteria:
- [x] Possiblity to set a deadline (optional!). After this deadline the students may not upload their answers or work on the questions.

felixbrucker:
https://github.com/angular/material2/issues/5648

felixbrucker:
As soon as `@angular/material-moment-adapter` hits npm we should use it.

See https://github.com/angular/material2/issues/7464

d89:
looks good 👍

PR #407, opened 2017-11-30
Use international date format for datepicker

felixbrucker:
## Improvements
- The date is now formatted internationally and time is included always in 'HH:mm' format

related to #50


felixbrucker:
I don't think so, you set dates, and 00:00 is the equivalent to the date. In fact that's exactly what the datepicker does, it just wasn't displayed previously

PR #420, opened 2017-12-05
show deactivated flag on course card

khase:
## Improvements
- shows the deactivate flag directly on the course card
image

## Known Issues:
_NONE_


app/webFrontend/src/app/course/course.component.html
```diff
@@ -8,27 +8,31 @@
...
       <mat-icon>mode_edit</mat-icon>
     </button>
 
-    <button mat-icon-button (click)="showReport(course._id)"
-            *ngIf="isCourseTeacherOrAdmin(course)">
+    <button mat-icon-button (click)="showReport(course._id)">
       <mat-icon>show_chart</mat-icon>
     </button>
 
+    <button mat-icon-button (click)="showReport(course._id)"
```
aeimer: Why is there a click-binding?

grosspersky: The other buttons of this component seem to have the same problem.
It's a bit strange why we aren't using the router here at all.

aeimer: No not the same problem! It makes no sense to add click here, the btn is only a flag and it is disabled anyway...

PR #439, opened 2017-12-14
Feature/240 add better error messages for registration page

### LINKED ISSUE #240: Add better error messages for registration page
# User Story

### As a: 
User

### I want:
more feedback when a student creates his account

### so that:
the student knows what is wrong with the input data

## Acceptance criteria:
- [x]  show individual error messages that describe what exactly is wrong. Not one general error message.
- [x] password should be at least 8 letters and one character that is not a letter.

alexweinfurter:
## Description:
Adds validation / hint to the teacher registration frontend which tells the user to use his @h-da.de email address. 
Besides fixes a bug that prevents the user to register as a teacher without a matriculation number.

Closes #240 

## Improvements
- Better user feedback
- Registration as a teacher without matriculation number works

## Known Issues:
_NONE_

------


app/webFrontend/src/app/auth/register/register.component.html
```diff
@@ -48,6 +48,7 @@
           <small *ngIf="registerForm.controls.email.hasError('email') && !registerForm.controls.email.hasError('required')" class="text-danger">Please use an email with the
             format john@doe.com
           </small>
+          <small *ngIf="registerForm.controls.email.hasError('hdaEmailValidator') && !registerForm.controls.email.hasError('email') && !registerForm.controls.email.hasError('required')" class="text-danger">Please register with your @h-da.de email</small>
```
PatrickSkowronek: Please replace the "@h-da.de" mail-ending with the value from the config

app/webFrontend/src/app/auth/register/register.component.ts
```diff
@@ -100,8 +102,24 @@ export class RegisterComponent implements OnInit {
...
       }),
-      email: ['', Validators.compose([Validators.required, Validators.email])],
-      uid: [null, Validators.required],
+      email: ['', Validators.compose([Validators.required, Validators.email, this.hdaEmailValidator.bind(this)])],
+      uid: ['', [this.uidValidator.bind(this)]]
     })
   }
+
+  hdaEmailValidator(control: FormControl) {
+    if (this.role === 'teacher' && !(control.value as String).includes('@h-da.de')) {
```
PatrickSkowronek: The "@h-da.de" should not be hard coded!
The function name should be different too.

khase: in my opinion it dosen' make sense to check this in the frontend. The frontend schould only check for a valid E-Mail and everything else should be verified by the API where it is configureable.

Just show the error messages provided by the api. (remenber to use the error codes)

errorCode (for example look at `app\webFrontend\src\app\auth\register\register.component.ts`):
```typescript
    noTeacher: {
      code: 'no teacher',
      text: 'You are not allowed to register as teacher'
    },```

**Comment by aeimer** [comment]
@alexweinfurter @khase why is this "simple" task still WIP? Can anyone else do this?

File: app/webFrontend/src/app/auth/register/register.component.ts (hunk 2 above)
Comment (alexweinfurter) [reply app/webFrontend/src/app/auth/register/register.component.ts]: I have built a service to use the teacherMailRegex that is set by the environment variables in backend. So the frontend can use the same regex as the backend does and the user gets feedback immediately while typing his email. I removed the hard coded @h-da.

File: api/src/controllers/AuthController.ts [hunk 3]
```diff
@@ -9,9 +9,9 @@ import emailService from '../services/EmailService';
 import {IUser} from '../../../shared/models/IUser';
 import {IUserModel, User} from '../models/User';
 import {JwtUtils} from '../security/JwtUtils';
-import config from '../config/main';
 import * as errorCodes from '../config/errorCodes'
 import {Course} from '../models/Course';
+import config from '../config/main'
```
lukaskorte: Why remove the semicolon?

api/src/controllers/ConfigController.ts
```diff
@@ -26,6 +26,15 @@ function publicConfig(id: string) {
 
 @JsonController('/config')
 export class ConfigController {
+
+  @Get('/env/:var')
```
lukaskorte: This is not a good design, because it requires a request for every environment variable.

We already have APIInfoController, which would be a better place for this.

alexweinfurter: I've fixed that.

PR #445, opened 2017-12-18
Feature/191 automatic enrollment

### LINKED ISSUE #191: Automatically enroll whitelisted users for the course they are whitelisted for
no text :).

OliverNeff:
## Description:
This feature should remove all students from whitelist when there are not whitelistet anymore, that means if a new CSV-File is uploaded and let a new regstration automatically add to coursers where there are whitelistet for. 

Closes #191 

## Improvements
- Optimize ObsCsvController.ts on some parts so there is no loading of all users needed anymore.
- Add await and async syntax on some methodes.
- new registrations are add to courses if there are whitelisted for it.

## Known Issues:
#191 

## Help on Line 15 in WhitelistUpdater needed. 
Want to use $elemMatch for finding alle courses where a user is whitelisted but it is not working. 
```typescript
    const courses = await Course.find(
      {
        whitelist:
          {
            $elemMatch:
              {
                firstName: user.profile.firstName.toLowerCase(),
                lastName: user.profile.lastName.toLowerCase(),
                uid: user.uid
              }
          }
      });
```

Should normally work.

------


OliverNeff:
It seems that there are not the same environments by tests local and in travis?

api/test/integration/auth.ts
```diff
@@ -52,6 +56,63 @@ describe('Auth', () => {
...
+        Course.create({
+          name: 'Test Course 1',
+          enrollType: 'whitelist',
+          whitelist: []
+        }).then((noElemCourse) => {
+          Course.create({
+            name: 'Test Course 2',
+            enrollType: 'whitelist',
+            whitelist: [whitelistUser]
+          }).then(elemCourse => {
```
lukaskorte: Pleaaase fix the indentation ... 😱

api/src/controllers/ObsCsvController.ts
```diff
@@ -83,22 +86,16 @@ export class ObsCsvController {
...
-      if (foundUsers.length > 0) {
-        foundUser = foundUsers[0];
-      }
-      if (foundUser != null && course.students.filter((student: any) =>
-        student.toString() === foundUser._id.toString()).length <= 0) {
-        course.students.push(foundUser);
+  private async addParsedUsersToCourse(course: ICourseModel) {
+    await Promise.all(this.whitelistUser.map(async (wUser) => {
+      const user: IUserModel = await User.findOne({uid: wUser.uid});
+      if (user && user.profile.firstName.toLowerCase() === wUser.firstName.toLowerCase() &&
```
lukaskorte: Do we really need to check the name when the uid matches?

api/src/utilities/WhitelistUpdater.ts
```diff
@@ -0,0 +1,28 @@
+import {IUser} from '../../../shared/models/IUser';
+import {Course} from '../models/Course';
+
+export class WhitelistUpdater {
```
lukaskorte: Why is this a class?

api/src/controllers/AuthController.ts
```diff
@@ -54,10 +57,13 @@ export class AuthController {
         return newUser.save();
       })
       .then((savedUser) => {
-        return emailService.sendActivation(savedUser)
-          .catch(() => {
-            throw new InternalServerError(errorCodes.errorCodes.mail.notSend.code);
-          });
+        // User can now match a whitelist.
+        return this.updater.addWhitelistetUserToCourses(savedUser).then(() => {
+          return emailService.sendActivation(savedUser)
```
lukaskorte: Please do not nest promises. Alternatively you could also introduce async/await here.

api/src/controllers/ObsCsvController.ts
```diff
@@ -37,9 +39,10 @@ export class ObsCsvController {
    * @param allUsers All users in System.
    * @returns {any} An updated course model.
    */
-  public updateCourseFromBuffer(buffer: string, course: ICourseModel, allUsers: any[]): ICourseModel {
+  public async updateCourseFromBuffer(buffer: string, course: ICourseModel) {
     this.pushWhitelistUser(buffer);
-    course = this.addParsedUsersToCourse(course, allUsers);
+    course.students = [];
+    course = await this.addParsedUsersToCourse(course);
```
lukaskorte: You are already passing a reference, no need to reassign.

api/src/controllers/ObsCsvController.ts
OliverNeff: It is agreed that uid, first name  and last name must match but it is checked by creation so you are right.

api/src/utilities/WhitelistUpdater.ts
OliverNeff: Because technically whitelists are technically not a thing of authentication, but I can move it to it.

api/src/controllers/AuthController.ts
OliverNeff: Hope it does the same as before. ^^

api/test/integration/auth.ts
OliverNeff: What is with call done on this way?

PR #446, opened 2017-12-19
Bugfixes mat fab btn on edit course content

aeimer:
## Description:
Makes the mat fab btn have more space on course edit.

------

### DIFF
--- a/app/webFrontend/src/app/course/course-edit/course-manage-content/course-manage-content.component.html
+++ b/app/webFrontend/src/app/course/course-edit/course-manage-content/course-manage-content.component.html
@@ -104,6 +104,8 @@
                     actionLabel="Add"></app-lecture-form>
 </div>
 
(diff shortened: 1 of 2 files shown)


aeimer:
@felixbrucker please check if this bug needs to be fixed or if it disappeared with your new course edit routing. 

Please give feedback to this pr :)

felixbrucker:
The new routing will not change this, so you are encouraged to merge this

lukaskorte:
This kind of "spacer" div is not good style. This looks like it can be solved using pure CSS.

Also the PR does not clearly state the issue that it is trying to solve.

aeimer:
There was no issue, @grosspersky handed the issue to me.

The Problem here was that the tab-content has overflow hidden. And even with a high z-index I was not able to move the button further to the bottom. So I had to add that spacer div. I know it is hacky, but it works...

PR #460, opened 2018-01-07
test import export duplicate

khase:
## Improvements
Implements tests for export, import and duplication of units, lectures and courses


aeimer:
@khase there is a conflict :/

aeimer:
The API seems to have a corrupt package-lock.json

PR #448, opened 2017-12-21
Course MediaManager - ALL

aeimer:
## Description:
Adds a MediaManager per Course

Closes #265 
Based on #466 __(merge this PR after #466)__

## Improvements
- Adds tab "Course edit/MediaManager"
- Introduces Mediamanager to Course edit

## Known Issues:
- Need to remove all old File-System

------
_Remember to prefix your PR-Title with `⚠ WIP: ` if you still work on it.
If you have reached a final state, remove the prefix._

Issue #265, opened 2017-10-26
Add Mediamanager per Course


aeimer:
# User Story

### As a: 
Teacher

### I want:
to manage my media (images, videos, other filetypes) per Course

### so that:
I have to upload a file once and can use it in that course by reference

## Acceptance criteria:
comming soon
- [x] add Tab to course "Mediamanager"

## Additional info:
Helpful modules could be:
- https://www.npmjs.com/package/angular-media-preview

utetrapp:
Each media belongs to a course. I may access media from another course and copy it into the current course (copy, no reference). Within a course I may add media from my (course) media management into a lecture. The link will be a hash based on the filename, so that I will be able to overwrite/update an existing file. If I delete a file from my media management, the system checks for existing references, warns/informs me, asks for confirmation and automatically removes the link in the content, i.e. the student should never get an missing media error.

PR #466, opened 2018-01-11
⚠ WIP: Course MediaManager - Backend

khase:
## Description:
This adds the backend for the MediaManager

related #448 
related #265 

## Improvements
- adds media Controller
- manage all media files in a folder hirarchie
- manages media per course

## Known Issues:


api/src/models/units/FileUnit.ts
```diff
@@ -12,56 +10,14 @@ interface IFileUnitModel extends IFileUnit, IUnitModel {
...
     }
   ],
   fileUnitType: {
     type: String,
     required: true
   }
-}, {
-  toObject: {
-    transform: function (doc: any, ret: any) {
-      ret._id = ret._id.toString();
```
khase: wouldn't it be useful to keep this transformation to convert the `_id` and the reference ids when they don't get populated?

grosspersky: The ``_id`` should be converted by the still existing method on the unit base schema.
The files can be handled after population works. We just reapply it then in a modified version.

aeimer:
Merged migration and BE into the FE branch/PR

PR #477, opened 2018-01-16
Now using generics for Abstract DataService functions

khase:


aeimer:
@grosspersky #478 is merged :)

aeimer:
Because I am using the readSingleItem funtion directly. I asume, I would break a lot of stuff if I'd apply this to all services....

aeimer:
@lukaskorte merged dev into

PR #482, opened 2018-01-17
Enhancement/use types for fixtureutils

khase:
## Description:

As requested by @OliverNeff all FixtureUtils should return proper Datatypes.

## Improvements
- FixtureUtils now return proper Datatypes instead of ``any``


aeimer:
@khase how can I test this?

khase:
just check if these functions return what they claim to?

The Tests and fixtures (which rely on all of thesefunctions) work well, so at least these changes didn't break anything

aeimer:
@khase do you get the error too?

PR #493, opened 2018-01-20
Bugfix/whitelist-synchronisation

OliverNeff:
## Description:
Syncronize users by adding and deleting whitelist users.
Users are now added  to cours if a whitelist user is created or delete from course if delete a whitelist user.

## Known Issues:
Review from 2018-01-18
see #501 
------

### DIFF
--- a/api/fixtures/FixtureUtils.ts
+++ b/api/fixtures/FixtureUtils.ts
@@ -8,6 +8,7 @@ import {IUnit} from '../../shared/models/units/IUnit';
 import {IUser} from '../../shared/models/IUser';
 import * as mongoose from 'mongoose';
 import ObjectId = mongoose.Types.ObjectId;
+import {IWhitelistUser} from '../../shared/models/IWhitelistUser';
 
 export class FixtureUtils {
   public static async getRandomUser(hash?: string) {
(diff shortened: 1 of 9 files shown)


OliverNeff:
Sorry I deleted the comment from coveralls: 
Coverage increased (+0.5%) to 62.802%

PatrickSkowronek:
I looked at the code, which is good. But I did not understand the logic behind the description of this pr. I can still add multiple times the same student to a whitelist. Another problem is uploading a csv file, which is like the test resource example. The first name and second name is switch:
screenshot from 2018-01-23 00-06-36

As a Student in the whitelist I cloud not enter a course. I did not see it either. Should that not be possible?

OliverNeff:
@PatrickSkowronek Have you updated your fixtures bevore you test it? 
This commit should prevent duplicates: 
https://github.com/h-da/geli/pull/493/commits/35af85610dcd97470c29780da9d14660009cdde2

see also stackoverflow: 
https://stackoverflow.com/questions/14692726/unique-constraint-with-two-fields-in-mongodb

PatrickSkowronek:
@OliverNeff I did updated the fixtures. But still the csv input is broken or saves the data wrong. I cloud add a Student which was already whitelisted with the csv file. In this case it did not work proper. I think this should be checked in the front and backend.

OliverNeff:
Then I look after it this evening again. You should enter the course when are whitelistet for it. This was implemented last semester.

OliverNeff:
Close because of #501

PR #494, opened 2018-01-21
Show access key and make better description for whitelist

OliverNeff:
## Description:
By review on 2018-01-18 we noticed that the access key is replaced with **** , but so you can't see the access key when you forgott it.
We decide that it is ok when it is shown in plane text to begin. 

## Improvements
- Access key is shown.

## Known Issues:
see #501 

------


app/webFrontend/src/app/course/course-edit/general-tab/general-tab.component.html
```diff
@@ -26,7 +26,8 @@
   </mat-form-field>
 
   <div class="csv-upload" *ngIf="enrollType === enrollTypeConstants.ENROLL_TYPE_WHITELIST">
-    <p>Upload your OBS csv file with your course members for whitelisting students.</p>
+    <p>Upload your OBS csv file with your course members for whitelisting students.
```
PatrickSkowronek: This should not be in this pr!

app/webFrontend/src/app/course/course-edit/general-tab/general-tab.component.ts
```diff
@@ -66,7 +66,7 @@ export class GeneralTabComponent implements OnInit {
         (val: any) => {
           this.course = val.name;
           this.description = val.description;
-          this.accessKey = val.hasAccessKey ? '****' : '';
+          this.accessKey = val.accessKey;
```
PatrickSkowronek: val.accessKey is undefined and the input field will be empty!

OliverNeff: Ohh sorry there is missing something in this merge request. ^^

app/webFrontend/src/app/course/course-edit/general-tab/general-tab.component.html
OliverNeff: You are right, it should be in: #493

app/webFrontend/src/app/course/course-edit/general-tab/general-tab.component.ts
OliverNeff: It should be done with this: 
https://github.com/h-da/geli/pull/494/commits/0c69e326696abb3c10312db02d3252b6f092a041

PR #524, opened 2018-03-18
Feature/265 add mediamanager per course migration

grosspersky:
## Description:

This provides migration for file units to work with the new media manager.


aeimer:
I will try to test it locally with some data from live

aeimer:
I try to get a live dump and share it with you. So you can test your code too

Issue #513, opened 2018-01-27
🐛 BUG: Can not  call error.json() method by all errors.


OliverNeff:
# User Story

### As a: 
User

### I want:
see meaningful error messages.

### so that:
I know where is an error.

## Acceptance criteria:
- [ ] There is not an error when an exception is thrown.
- [ ] Error messages are shown well. 
- [ ] Duration of snackbar is long enough that you can read the message.
- [ ] You can click on dismiss by all error messages.

## Additional info:

------
_Please tag this issue if you are sure to which tag(s) it belongs._

PatrickSkowronek:
The behavior of httperror from angular changed with angular5. This should be done with #572 !

PR #526, opened 2018-03-21
Bugfix/#520 fix showing full error message

### LINKED ISSUE #520: BUG: Fix Error Message showing full http 400 Message in Change Password Screen
### As a: 
User

### I want:
to see the Right Error Message, if i enter my PAssword wrong.

### so that:
I know what i did wrong.

## Acceptance criteria:
- [x]  Error Message in Change PAsswort Screen does not show the full http 400 Error, but only the thrown backend message

HPunktOchs:
## Description:
The Full ErrorMessage won't be displayed in the Frontend, but only the thrown Message on a failed attempt.

Closes #520 

## Known Issues:
- might be related to #513


aeimer:
Works if the API is alive with wrong password. Does not work if the API is down

PR #535, opened 2018-03-24
Bugfix api doc build

aeimer:
## Description:

Closes #

## Improvements
- 

## Known Issues:
_NONE_

------

### DIFF
--- a/api/package.json
+++ b/api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "geli-api",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "license": "GPL-3.0",
   "repository": "h-da/geli",
   "description": "This is the API-Package.json for geli.",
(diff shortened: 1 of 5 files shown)


micpah:
Looks like the root cause is that `npm prune --production` is called within docker.sh which is run prior to apidoc.sh. This removes all dev depenencies including all gulp packages, right?

api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "geli-api",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
micpah: Why not 0.5.1? And you also need to adjust both package-lock.json files.

aeimer: Because the comming version contains not only bugfixes, but also the media manager. According to semver we need to increase the minor version not the bugfix version.

PR #625, opened 2018-04-14
WIP:fix: handle promise rejection in getUser

### LINKED ISSUE #204: 🐛 BUG: Unhandeled Promise Rejection in User tests.
``` 
 User
    GET /api/users
      ✓ should return all users
      ✓ should fail with wrong authorization
    GET /api/users/roles
      ✓ should fail with wrong authorization
      ✓ should return an array with the defined roles
    PUT /api/users
(node:7360) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Can't set headers after they are sent.
(node:7360) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
      ✓ should fail with bad request (admin privileges)
      ✓ should fail with bad request (email)
      ✓ should fail with wrong authorization (role edit)
      ✓ should fail with wrong authorization (uid)
      ✓ should update user data
```

armel22:
- return status code 404 if user not found
- use regex on id parameter to avoid /users/roles matching /users/:id

## Description:

Closes #204 

## Improvements
- `/users/roles` doesn't match `/users/:id`  anymore.

## Known Issues:
_NONE_

------

### DIFF
--- a/api/src/controllers/UserController.ts
+++ b/api/src/controllers/UserController.ts
@@ -240,6 +240,8 @@ export class UserController {
    *
    * @apiSuccess {User} user User.
    *
+   * @apiError NotFoundError User was not found.
+   *
    * @apiSuccessExample {json} Success-Response:
    *     {
    *         "_id": "5a037e6a60f72236d8e7c81d",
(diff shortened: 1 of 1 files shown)


PatrickSkowronek:
Should `getUser(@Param('id') id: string, @CurrentUser() currentUser?: IUser) ` be an async function.

PatrickSkowronek:
@armel22 last semester we changed our coding style from .then to await(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). In my opinion it is more clean then a ".then" promise chain. Also it should be same on all functions.

PatrickSkowronek:
@armel22 yeah only this function, all other is a bit overkill.

PR #627, opened 2018-04-15
Fixed restart of fileupload

### LINKED ISSUE #380: 🐛 BUG: can't restart a fileupload
when a fileupload fails you can't restart it by clicking upload again.
In order to restart it (or upload something else) you currently need to close and reopen the unit.

fwitulski:
Restart of file upload after cancellation or error is now possible

## Description:
added reset of fileUploader status variable "isUploading", so restart of file upload is possible

Closes #380 

## Improvements
- 

## Known Issues:
_NONE_

------

### DIFF
--- a/app/webFrontend/src/app/shared/components/upload-form/upload-form.component.ts
+++ b/app/webFrontend/src/app/shared/components/upload-form/upload-form.component.ts
@@ -1,6 +1,7 @@
 import {Component, EventEmitter, Input, OnChanges, OnInit, Output, SimpleChanges} from '@angular/core';
 import {FileUploader, FileUploaderOptions} from 'ng2-file-upload';
 import {MatSnackBar} from '@angular/material';
+import {errorCodes} from '../../../../../../../api/src/config/errorCodes';
 
 @Component({
   selector: 'app-upload-form',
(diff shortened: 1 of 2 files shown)


app/webFrontend/src/app/shared/components/upload-form/upload-form.component.ts
```diff
@@ -60,6 +61,7 @@ export class UploadFormComponent implements OnInit, OnChanges {
 
     this.fileUploader.onCancelItem = () => {
       this.fileUploader.isUploading = false;
+      throw(errorCodes.upload.cancel);
```
PatrickSkowronek: You throw a new error, but what I expected was that there should not be any error when you cancel a upload, or when the upload fail. There is properly a try and catch missing.

fwitulski: okay then I missunderstood your previous comment. Will have to take a deeper look, because the cancel function comes from an external library.

fwitulski:
Behaviour of the upload slightly changed. If you cancel a file now the next file is automatically processed, beforehand the whole upload was "paused". Is that fine?

PatrickSkowronek:
I think that it is fine, i can't find a good reason not to stop the upload of multiple files. @d89 what do you think. What is the behavior of moodle? Should we add a cancel all button.

PatrickSkowronek:
It's working as expected if @d89 want a "cancel all" button, we should open a new issue!

PR #634, opened 2018-04-17
⚠ WIP_Bugfix/#581 profile pic upload

### LINKED ISSUE #581: Upload of profile picture currently accepts files of all types
### Expected Behaviour
The Upload component should not accept files of types that never could be represented in the profile section. Maybe it would be even wise to block files that are of unnecessary sizes.  
### Actual Behavior
Currently all data-types (.jpg, .json, .zip, text-files and even an iso) will be accepted in the file upload of the Edit-Profile screen. Regardless of the size (even +4GB) the upload will be started and the "picture" will then be displayed (which currently fails) in the profile picture section. 

## Acceptance criteria:
- [ ] Add a filter to upload component
------

MichaelNarkus:
## Description:
In the webfrontend it is now forbidden to choose non-images to upload.
The UserController now checks, wether the uploaded file is of the mimeType-Family 'image' and will throw an ForbiddenError, when the file is not an image

Closes #581

## Improvements
- Other Files ( f.e. .zip or .exe) now can no longer be uploaded and displayed as a profile picture

## Known Issues:
_NONE_

------


app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
```diff
@@ -2,7 +2,7 @@
   <mat-tab label="Upload image">
     <div class="tab-content">
       <!-- TODO: Replace with shared upload form -->
-      <input type="file" name="fileupload" ng2FileSelect [uploader]="uploader" placeholder="Upload picture"/>
+      <input type="file" name="fileupload" ng2FileSelect [uploader]="uploader" placeholder="Upload picture" accept="image/*"/>
```
PatrickSkowronek: If there is an error it will not be catched!

MichaelNarkus: What specific error are you referring to? The highlighted codeline only changes the files that can be chosen within the opened file-picker. If needed I could also check the extension of the upload file before posting. 
If a non-image will be sent to the backend (f.e. via post), node, with the changes in the user-controller, will reject it with a ForbiddenError. On my end the current image will not be deleted, when the upload will be cancelled  or throws the exception. Can you please send me more details about the file you were uploading, so I can try to reproduce this?

api/src/controllers/UserController.ts
```diff
@@ -10,6 +10,7 @@ import {isNullOrUndefined} from 'util';
 import {errorCodes} from '../config/errorCodes';
 
 const multer = require('multer');
+const mime = require('mime-types');
```
grosspersky: You don't need this to check for the mime type.

api/src/controllers/UserController.ts
```diff
@@ -318,14 +319,20 @@ export class UserController {
...
-        user.profile.picture = {
-          _id: null,
-          name: file.filename,
-          alias: file.originalname,
-          path: file.path,
-          size: file.size
-        };
-        return user.save();
+        // look up mime-type of uploaded file (f.e. a jpg will return 'image/jpg'
+        const mimeFamily = mime.lookup(file.originalname).split('/', 1)[0]; // split received string at '/' to return family
```
grosspersky: Just use ``file.mimetype`` for this check.

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
grosspersky: You shouldn't use the ``accept`` attribute of the input type file. Instead you should configure the uploader to match the possible files.
But it would be much better if you incorporated the shared file upload form like the TODO suggest. This component has an interface that allows you to specify the mime types and it allows you to upload files using drag and drop.

api/src/controllers/UserController.ts
MichaelNarkus: Thanks! I tried using file.mimetype() before, which obviously didn't work

PatrickSkowronek:
As @grosspersky pointed out, you still can change in the file upload the list to any and upload a file. Than the backend send a httperror, and the frontend ignores it and make a positive snackbar. This should be avoided.
peek 2018-04-18 22-03
If you still upload something bad, there should be a negative snackbar.

PR #637, opened 2018-04-18
 #597: added email validator to reset form, show hint if the email isn…

### LINKED ISSUE #597: 🐛 BUG: Validate password reset email
Currently we do not validate the e-mail in the frontend and backend.

- [x] Add empty filter, so nobody can send an empty e-mail address
- [ ] Add e-mail filter set in config (Optimal)

lukas-schardt:
…'t validated correctly

## Description:

Closes #597

## Improvements
- Passwort reset validates

## Known Issues:
_NONE_

------


app/webFrontend/src/app/auth/reset/reset.component.ts
```diff
@@ -80,7 +85,7 @@ export class ResetComponent implements OnInit {
       this.resetForm = this.formBuilder.group({});
     } else {
       this.resetForm = this.formBuilder.group({
-        email: ['', Validators.required]
+        email: ['', Validators.email ]
```
PatrickSkowronek: The Validation does work properly. A email with without "@" does throw a error. A email with missing domain ending, like ".de", will accept. This should also fail.

lukas-schardt: An email can still be valid without a top level domain: see https://en.wikipedia.org/wiki/Email_address#Examples 
But if it really is required to only support emails with tld, we would need to write a custom validator and couldn't use the default angular email validator

PatrickSkowronek: @lukas-schardt interesting fact, didn't know that, but it is prohibited to used those since 2013, so would propose using a custom validator.

app/webFrontend/src/app/shared/validators/validators.ts
```diff
@@ -12,3 +12,20 @@ export function matchPasswords(passwordKey: string, confirmPasswordKey: string)
...
+/**
+ * A custom email validator which, in comparison to the default angular email validator,
+ * also checks if the email address has a top-level domain.
+ *
+ * @param {FormControl} formControl
+ * @returns {{emailValidator: {valid: boolean}}}
+ */
+export function emailValidator(formControl: FormControl) {
+  // tslint:disable-next-line:max-line-length
+  const EMAIL_REGEX = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
PatrickSkowronek: Please Add the link to the source!

PR #656, opened 2018-04-20
Provide valid configuration for ng test

### LINKED ISSUE #127: Activate Angular-Unit-Tests
# User Story

### As a: 
Developer

### I want:
to be able to run `ng test`

### so that:
the angular-components gets tested

## Acceptance criteria:
- [x] Testing is possible
  - [x] on local dev-machine
  - [x] in Travis

### LINKED ISSUE #655: Configuration for ng test broken
# User Story

## Acceptance criteria:
- [x] Provide working configuration for ng test

## Additional info:

image

danielkesselberg:
## Description:

Closes #655 
Closes #127 (including travis)

image

## Improvements
- ng test compiles source and run frontend tests
- ng test is running at travis

## Known Issues:
- most of the frontend tests are broken
- when ng test fails the build does not fail (i did this on purpose because most tests are broken)


aeimer:
@PatrickSkowronek why have you changed the `npm run lint` to `ng lint`? That's bad practice

danielkesselberg:
I changed it. The main reason are the additional arguments for ng test. I tried to avoid a new command like ci-test with these arguments in package.json. I dont see any problem by using ng xxx instead npm xxx. Basically the npm commands are only wrapper. So why do you call this bad practice?

PatrickSkowronek:
@danielkesselberg I found this https://github.com/angular/angular-cli/issues/967#issuecomment-223168638 which explains that `ng lint` just calls `npm run lint` . Even the angular-cli uses `ng lint` https://github.com/angular/angular-cli/blob/8e169b8a102d85511bc19eff7a8c3c2961facc31/tests/e2e/tests/lint/lint-no-config-section.ts . In my opinion it is correct. @aeimer why should it be bad practice.

aeimer:
> I found this angular/angular-cli#967 (comment) which explains that ng lint just calls npm run lint .

I don't know if this is for the dev-env of the ng-cli setup or as a general behaviour of ng lint. Anyways this comment is 2 years old...

> Even the angular-cli uses ng lint https://github.com/angular/angular-cli/blob/8e169b8a102d85511bc19eff7a8c3c2961facc31/tests/e2e/tests/lint/lint-no-config-section.ts .

Thats propably the easiest way for them, as they **develop** ng cli and they can run ng-commands directly from the code.

> The main reason are the additional arguments for ng test.

RTFM => https://docs.npmjs.com/cli/run-script#description

In my opinion we should use everywhere the `npm run [scriptname]` command, this way nobody is confused. We also can use `npm run` everywhere, also in the backend. We should provide a single command to run everything and we should run exactly the same commands in dev and on travis.

So I would have written something like:
```
- npm run lint        # run linter
- npm run test -- --single-run true --watch false        # run test - does not work properly
- npm run e2e         # run end-to-end-tests
```

danielkesselberg:
Sounds good to me. Im going to create another pr for this.

Issue #594, opened 2018-04-07
🐛 BUG: Severe CourseController information leak 🔥


torss:
Unauthorized information for all courses is transmitted to any logged in user.
For example, anyone can easily read **every** course access key using only the built-in browser dev tools.
This is a severe issue and should be fixed ASAP.

Tested access key vulnerability on the staging system using Chrome 65 and confirmed the leak on the live system.

torss:
Note that there is a related TODO in [CourseController.ts](../blob/develop/api/src/controllers/CourseController.ts)

> TODO: Do not send lectures when student has no access

but no mention of the "accessKey" property in the "Success-Response" documentation, so I assume this leak issue is unknown.

There is also another TODO in the same file for the "PostCourseWhitelist" API:

> TODO: Needs more security

This probably should get its own GitHub issue.

d89:
You would need to present how you fixed it and how it was possible beforehand to sneak the key. I'd say it looks good for now 👍

PR #664, opened 2018-04-22
Bugfix/#654 wasteful course data fix

### LINKED ISSUE #653: 🐛 BUG: CourseController getCourse leak 🔥
The `/api/courses/:id` getCourse API still leaks important information - such as every enrolled student's email address - to anyone who can view the course.
This issue is closely related to (and was missed in) #594.
It was discovered while working on #82.

### LINKED ISSUE #654: 🐛 BUG: Wasteful frontend course data requirements 🚯
Currently the frontend requires far more information than it should to display the dashboard.
E.g. for teachers, an array of the course's teachers is searched to determine whether or not the course belongs to the current user (see `filterMyCourses` in [dashboard-teacher.component.ts](../blob/develop/app/webFrontend/src/app/start/dashboard/dashboard-teacher/dashboard-teacher.component.ts)).

This worked in the past because the entire course information was leaked anyway, as noted in #594.
To fix those leaks in #594, without larger frontend changes, an overly complicated filter code was applied.

It would however be more sensible to properly minimize both the required data in the frontend, as well as the sent data in the backend.
The goal is thus to only send data that is actually necessary / intended for display in the frontend, without resorting to complex user-role-specific filtering.

At least changes of both `/api/courses/` getCourses and `/api/courses/:id` getCourse in [CourseController.ts](../blob/develop/api/src/controllers/CourseController.ts) are required, with accompanying frontend logic adjustments as needed:

- [x] Focused `/api/courses/` getCourses response & working frontend
- [x] Focused `/api/courses/:id` getCourse response & working frontend

torss:
## Description

Closes #654
Closes #653

This streamlines the `getCourses` and `getCourse` APIs to fix #654, also fixing the #653 `getCourse` leak in the process.

A severe leak in `getCourses` was previously fixed in #594 via an overly complex `getSanitized` filter. Such a filter was necessary, because the dashboard frontend required various unnecessary data to function properly. (E.g. the course teacher IDs to check whether a teacher is part of a course or not.)
This has been streamlined via a new `ICourseDashboard` interface, which specifies all required data for the dashboard. That data includes some of the original `ICourse` fields, but also some flags from the (improved) `checkPrivileges` course method. The frontend & backend has been adjusted accordingly and the now obsolete `getSanitized` functionality is removed.

More importantly, another leak (#653) in `getCourse` is fixed, which is done implicitly via the streamlining (#654). That streamlining is facilitated by splitting `getCourse` into `getCourseView` and `getCourseEdit`.
`getCourseEdit` continues to provide the "unsafe" (i.e. full) course information for editing purposes, but now only allows access for authorized users (i.e. admins or course teachers/admins).
`getCourseView` provides a reduced (new) `ICourseView` object, similar to `ICourseDashboard`, but with lecture information instead of the `checkPrivileges` access flags.
The API tests have been adjusted and extended to test against unauthorized users.

There are also other minor improvements / additions, mostly to support the primary goals.

## Improvements

- Refactored `checkPrivileges` course method
- New `ExtractId.ts` utility (to extract mongoose IDs)
- `ExtractId.ts` utility tests
- Simplified `getCourses` via `ICourseDashboard`
- Split `getCourse` into `getCourseView` and `getCourseEdit`
- Secured `getCourseEdit` via access control
- Secured `getCourseView` via reduction to `ICourseView`
- Moved some functionality from `getCourse(s)` to `Course.ts` / `Lecture.ts`
- Removed obsolete `getSanitized` course method
- Adjusted frontend to the streamlined / secured `getCourse(s)` system
- Adjusted tests to the streamlined / secured `getCourse(s)` system
- Minor typo fixed in `Access-key-test.json`
- Import clean-up in `CourseController.ts`

## Known Issues

Due to the amount of central (course) system changes, some parts might clash with concurrent work.


app/webFrontend/src/app/start/dashboard/dashboard.component.ts
```diff
@@ -26,12 +26,12 @@ export class DashboardComponent implements OnInit {
   }
 
   getCourses() {
-    this.courseService.readItems<ICourse>().then(courses => {
+    this.courseService.readItems<ICourseDashboard>().then(courses => {
```
PatrickSkowronek: I would change it to async, it will be much better!

app/webFrontend/src/app/course/course-detail/course-detail.component.ts
```diff
@@ -39,8 +39,8 @@ export class CourseDetailComponent implements OnInit {
   }
 
   getCourse(courseId: string) {
-    this.courseService.readSingleItem(courseId).then(
-      (course: any) => {
+    this.courseService.readCourseToView(courseId).then(
```
PatrickSkowronek: This should be also async and await

api/src/controllers/CourseController.ts
```diff
@@ -138,129 +113,229 @@ export class CourseController {
...
-      lecture.units = await Promise.all(lecture.units.map(async (unit: IUnitModel) => {
-        unit = await unit.populateUnit();
-        return unit.secureData(currentUser);
-      }));
-
-      return lecture;
-    }));
-
+    await course.processLecturesFor(currentUser);
     return course.toObject();
```
grosspersky: Couldn't you chain the toObject() method here too, since you are returning the mongoose object itself in the procress function?

api/src/controllers/CourseController.ts
```diff
@@ -138,129 +113,229 @@ export class CourseController {
...
+    // This is currently a redundant check, because userReadConditions in getCourse above already restricts access!
+    // (I.e. just in case future changes break something.)
+    if (!course.checkPrivileges(currentUser).userCanViewCourse) {
+      throw new ForbiddenError();
+    }
+
+    await course.populateLecturesFor(currentUser)
+      .execPopulate();
+    await course.processLecturesFor(currentUser);
+    return course.forView();
```
grosspersky: See next comment.

grosspersky: See previous comment.

api/src/utilities/ExtractId.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * Tries to extract the id from a mongoose object.
+ * This could be a node-mongodb-native ObjectID (i.e. mongoose.Types.ObjectId), via toHexString,
+ * or anything that contains an 'id' property, which is directly returned.
+ *
+ * @param from A mongoose object.
+ * @param fallback Return this if no id is found.
+ */
+function extractIdImpl(from: any, fallback?: any) {
```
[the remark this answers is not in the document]
torss: I don't think that the name really needs a "...FromObject" part, but yes, a more specific name for the file & functions might be sensible. How about `extractMongoId` for brevity?

danielkesselberg:
I guess extractMongoId will do it.

Am 23. April 2018 22:35:02 MESZ schrieb Torsten Schlett <notifications@github.com>:

app/webFrontend/src/app/course/course-detail/course-detail.component.ts
torss: Hmm, I'm not sure how to elegantly deal with the error handling here when using async / await.
The original function body has 11 LOC _(Just FYI; I don't intend LOC to be a decisive factor.)_.

Await variant 1 ((probably completely) equivalent to the original code), 10 LOC:

```ts
try {
    this.course = await this.courseService.readCourseToView(courseId);
} catch (errorResponse) { // NOTE https://github.com/Microsoft/TypeScript/issues/20024
    if ((<Response>errorResponse).status === 401) {
        this.snackBar.open('You are not authorized to view this course.', '', {duration: 3000});
    }
    return;
}
LastVisitedCourseContainerUpdater.addCourseToLastVisitedCourses(courseId, this.userService, this.userDataService);
this.titleService.setTitleCut(['Course: ', this.course.name]);
```

Await variant 2 (NOT equivalent to the original), 9 LOC:

```ts
try {
    this.course = await this.courseService.readCourseToView(courseId);
    LastVisitedCourseContainerUpdater.addCourseToLastVisitedCourses(courseId, this.userService, this.userDataService);
    this.titleService.setTitleCut(['Course: ', this.course.name]);
} catch (errorResponse) { // NOTE Here errorResponse could be something thrown after the await line!
    if (errorResponse.status === 401) {
        this.snackBar.open('You are not authorized to view this course.', '', {duration: 3000});
    }
}
```

Await variant 3 (functionally equivalent to the original), 8 LOC (+1 with new import):

```ts
// NOTE Requires https://github.com/scopsy/await-to-js (or we could just copy the tiny code)
let errorResponse: Response;
[errorResponse, this.course] = await to(this.courseService.readCourseToView(courseId));
if (!errorResponse) {
    LastVisitedCourseContainerUpdater.addCourseToLastVisitedCourses(courseId, this.userService, this.userDataService);
    this.titleService.setTitleCut(['Course: ', this.course.name]);
} else if (errorResponse.status === 401) {
    this.snackBar.open('You are not authorized to view this course.', '', {duration: 3000});
}
```

So IMO we should either go for variant 3 and add the tiny dependency - or just keep the original.

app/webFrontend/src/app/start/dashboard/dashboard.component.ts
torss: Agreed!

api/src/controllers/CourseController.ts
```diff
@@ -62,72 +59,50 @@ export class CourseController {
...
+   *             "userCanViewCourse": false,
+   *             "userIsCourseAdmin": false,
+   *             "userIsCourseTeacher": false,
+   *             "userIsCourseMember": false
    *         }
    *     ]
    */
   @Get('/')
-  async getCourses(@CurrentUser() currentUser: IUser) {
+  async getCourses(@CurrentUser() currentUser: IUser): Promise<ICourseDashboard[]> {
```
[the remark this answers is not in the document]
torss: Are you sure about the return type being incorrect?
I think async functions naturally (have to) return a `Promise`.

api/src/controllers/CourseController.ts
torss: Hm, yes, that would look like
```ts
return (await course.processLecturesFor(currentUser)).toObject();
```
but I think keeping the lines separate is better - do you agree?

danielkesselberg:
You are right. I've been messing things up.

Am 24. April 2018 00:36:30 MESZ schrieb Torsten Schlett <notifications@github.com>:

api/src/controllers/CourseController.ts
grosspersky: The correct way would be
```
return course.processLecturesFor(currentUser).toObject();
```
You shouldn't need the await here, since ``toObject()`` would be called on the result of ``procressLecturesFor()``, which needs to finish first anyway.

But you could evaluate if this is really the case or if I remember something wrong.
I just think it would look cleaner if the methods are chained and the then unnecessary ``await`` is gone. But as I said, only if this results in the expected behaviour.

api/src/controllers/CourseController.ts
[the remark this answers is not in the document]
danielkesselberg: You are right. I've been messing things up.

app/webFrontend/src/app/course/course-detail/course-detail.component.ts
PatrickSkowronek: 2 and 3 are good solutions, I would do 2, but it is up to you!

api/src/controllers/CourseController.ts
grosspersky: Yes, if it isn't possible just keep it in the current style.
Because of this I wrote you could ``evaluate`` if it is possible. If it isn't thats totally okay.

And the populateUnit()-Method was created by me. ;-)

torss: Yeah, I also wasn't sure about whether this is possible or not, thus the long-ish explanation. :)

app/webFrontend/src/app/course/course-detail/course-detail.component.ts
grosspersky: I would prefer variant 2 too. I think a visible try/catch-Block is much better to see that error handling is handeled at this place and we don't need another dependency here.

torss: But variant 2 is functionally flawed, it will capture all exceptions, even those below the `await` line (i.e. anything that might ever be thrown in `addCourseToLastVisitedCourses` or `setTitleCut`). In fact I actually only put it there to demonstrate the issue that refactoring with `await` implies, when error handling is involved.
_(And another minor inconvenience with try/catch in general is the lack of type support for `errorResponse`, at least in the TypeScript version we use.)_
Are absolutely sure about this?

torss: As decided in the meeting today, we go for variant 2 with some slight modifications (i.e. `try` line order change and an additional, general `else` error notification to prevent silent failures).

Issue #591, opened 2018-04-06
Wanted behaviour in Progress Overview in Profile?


fwitulski:
As a student u can submit the tests again after reloading the course page.
The course then registers those entries as negative nothing (-1) and either tried or done (+1) depending on test result.
As result u get large % numbers in the progress overview in your profile with negative numbers/ numbers larger than actucal test parts.

![grafik](https://user-images.githubusercontent.com/32933845/38447359-237a1838-39fd-11e8-81eb-c90515e63d2f.png)

Is this the wanted behaviour?
------
_Please tag this issue if you are sure to which tag(s) it belongs._

grosspersky:
The wanted behaviour should be and has been at one point, that your submitted progress gets displayed on the units if you reload the course.

Additional submission would also be possible but they should update the already existing progress object. This definitly isn't a wanted behaviour in this form, since all values should always make up 100%.

lukas-schardt:
Possibily related to #612

PR #681, opened 2018-04-25
⚠WIP: Bugfix/#612 course progress gone after reload

### LINKED ISSUE #612: 🐛 BUG: Course progress is gone after page reload
If I work on a course as a student and submit tasks, the progress is not visible to me anymore if I reload the page(also go to another page and continue working on the course later).

So I cannot see what I have done in the course before.
------

lukas-schardt:
## Description:

Closes #612, #591 

## Improvements
- The course progress is now restored after page reloads and multiple submits don't add each other up anymore

## Known Issues:

data.service.ts could use some refactoring (e.g. being able to pass an apiPath to the methods instead of having to change and reset the member variable)

## Questions:

Should the user code in the code-katas be saved in the progress table? Could lead to js-injection in the future.

------

### DIFF
--- a/app/webFrontend/src/app/shared/services/data.service.ts
+++ b/app/webFrontend/src/app/shared/services/data.service.ts
@@ -63,9 +63,9 @@ export abstract class DataService {
       .toPromise();
   }
 
-  readSingleItem<T>(id: string): Promise<T> {
+  readSingleItem<T>(id: string, apiPath = this.apiPath): Promise<T> {
     return new Promise((resolve, reject) => {
-      this.backendService.get(this.apiPath + id)
+      this.backendService.get(apiPath + id)
         .subscribe(
           (responseItems: any) => {
             if (this.changeProps2Date) {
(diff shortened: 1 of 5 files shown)


torss:
@lukas-schardt  Just a quick note re. GitHub usage:

> Closes #612, #591

You can't close multiple issues like that, see https://help.github.com/articles/closing-issues-using-keywords/#closing-multiple-issues for details.

app/webFrontend/src/app/shared/services/data.service.ts
```diff
@@ -64,9 +64,9 @@ export abstract class DataService {
       .toPromise();
   }
 
-  readSingleItem<T>(id: string): Promise<T> {
+  readSingleItem<T>(id: string, apiPath = this.apiPath): Promise<T> {
     return new Promise((resolve, reject) => {
-      this.backendService.get(this.apiPath + id)
```
PatrickSkowronek: We should discuss this @lukas-schardt . Why is it better than before?

lukas-schardt: @PatrickSkowronek There was a problem with async/await.
If we wanted to call the method a few times in a short period of time, the api path would concatenate. Will create a follow up ticket to refactor the data.service to work with an async/await sturcture

PR #690, opened 2018-04-26
Bugfix/581 picture upload

### LINKED ISSUE #581: Upload of profile picture currently accepts files of all types

MichaelNarkus:
## Description:
BackEnd and FrontEnd will now check the mimeTypes of uploaded images.
The Upload of a new profile picture now supports Drag&Drop via ng2-file-upload.

Closes #581 

## Improvements
- Backend will now reject nonimages
- Drag&Drop in uploadDialog

## Known Issues:
_NONE_

------


app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
```diff
@@ -1,12 +1,20 @@
...
-        Upload picture
-      </button>
-      <button mat-raised-button (click)="cancel()">Cancel</button>
+      <label ng2FileDrop
+             [ngClass]="{'nv-file-over': true}"
+             [uploader]="uploader"
+             class="file-dropzone"
+              (onFileDrop)="startUpload()">
+        <span>Drop files here or click here to add files</span>
+        <input type="file"
```
PatrickSkowronek: Is it possible to set "image" instead of "file"?

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.ts
```diff
@@ -121,4 +119,9 @@ export class UploadDialog implements OnInit {
     this.dialogRef.close(false);
   }
 
+  public startUpload() {
+    this.uploader.uploadAll();
```
PatrickSkowronek: Please catch the error if another file is trying to uploaded instead of a image!

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
```diff
@@ -1,12 +1,20 @@
 <mat-tab-group *ngIf="!showProgressBar" (selectChange)="changedTab($event)">
   <mat-tab label="Upload image">
     <div class="tab-content">
-      <!-- TODO: Replace with shared upload form -->
-      <input type="file" name="fileupload" ng2FileSelect [uploader]="uploader" placeholder="Upload picture"/>
-      <button mat-raised-button color="primary" disabled="{{!uploader.queue.length}}" (click)="uploadImage()">
-        Upload picture
-      </button>
-      <button mat-raised-button (click)="cancel()">Cancel</button>
+      <label ng2FileDrop
```
grosspersky: Now it is copy & paste from the upload form. Don't do this and use the shared upload form like this:
```
  <app-upload-form [uploadMethod]="'POST'"
                   [uploadPath]="uploadPath"
                   [additionalData]="null"
                   (onFileSelectedChange)="null"
                   (onFileUploaded)="onFileUploaded($event)"
                   (onAllUploaded)="onAllUploaded()"></app-upload-form>
```

Provide data needed data in the corresponding component.ts file. To use the form correctly take a look at available input and output parameters which allow for communication with the unit:
```
  @Input()
  uploadPath: string;

  @Input()
  uploadMethod = 'POST';

  @Input()
  allowedMimeTypes: string[];

  @Input()
  additionalData: any;

  @Input()
  maxFileNumber: number;

  @Output()
  onFileSelectedChange = new EventEmitter();

  @Output()
  onFileUploaded = new EventEmitter();

  @Output()
  onAllUploaded = new EventEmitter();
```

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
MichaelNarkus: Although input type="image" is a valid input, adding it here will result in the the application no longer opening the file picker at all. Only the DropZone will work with that.

MichaelNarkus: It's possible to add the accept="image/+" instead, which would suggest image file-types within the file picker.

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.ts
MichaelNarkus: As the uploader accepts only files of image-types, non-images will never be added to the queue. On my end no error will be thrown as the queue is empty.

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
PatrickSkowronek: That would be perfect!

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.ts
PatrickSkowronek: What is about the error which can come from the backend, if i upload a file that is not an Image?

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
MichaelNarkus: Added the suggested file type.

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.ts
MichaelNarkus: In the context of the front-end upload, that could not occur. The check in the backend is needed though in case a post-request would be sent directly to the backend (f.e. via Postman). I'll add a catch-phrase later today though, since it could be possible that other errors occur (bad-route or backend-down) while uploading.

app/webFrontend/src/app/shared/components/upload-dialog/upload-dialog.component.html
grosspersky: To prevent this circular dependency and since the dialog module is the most "inner" module that is using the upload form, you should move the declareration of the upload form to this module.

Then, to keep it available for other modules you should import and then rexport the component from the shared module.

PatrickSkowronek: @grosspersky at the moment I like the new form, switching from dialog to dialog is not a good idea. What do you think?

grosspersky: I'm not talking about switching between dialogs, which indeed isn't a good idea.
But at the moment we have an unwanted code replication between this dialog and the upload-form component.

You should note that the upload-form component is different from the upload-form-dialog component which was used in the last PR.

If the component is used you have the same behavious.

Also by testing I noticed some not so good design choices:
* The image is uploaded immediately after dragging it on the drop area. It would be better to still confirm the chosen image by a button click to upload it.
* The webcam upload depended on the regular upload so it needs to be verified that the upload of webcam pictures is still working.

MichaelNarkus: The webcam upload is indeed still working and has to rely on the existing fileuploader within uploadDialog.
Starting the upload directly when a file is selected and has the correct file type was a design decision I discussed with PatrickSkowronek as the maxFileNumber would be 1. Of course that can be discussed further.

PatrickSkowronek:
Bug is fixed, refactoring is will be done too #693 !

PR #738, opened 2018-05-13
Fix apidoc syntax

### LINKED ISSUE #737: Apidoc broken
Apidoc generation broken due syntax error

danielkesselberg:
## Description:

Closes #737 

## Improvements
- Apidoc generation fixed
- Generate Apidoc with travis and fail when syntax error


.travis.yml
```diff
@@ -72,6 +72,7 @@ script:
   - cd api
   - sed -i 's/$TRAVIS_COMMIT/'$TRAVIS_COMMIT'/' src/server.ts
   - npm run test        # run backend test
+  - npm run apidoc      # generate apidoc
```
PatrickSkowronek: This is a bad idea, I wrote a better script for it!

PatrickSkowronek:
@danielkesselberg what do you think about the script?

danielkesselberg:
@PatrickSkowronek looks fine. I added chmod +x and remove fi before elif/else. We may change the name to keep up with the other scripts (- instead of _). I would try to merge if/elif block like below but this only works for bash and is not portable:

`if [[ "$TRAVIS_PULL_REQUEST" != "false" && ( "$TRAVIS_BRANCH" == "develop"  || "$TRAVIS_BRANCH" == "master"  ) ]]; then`

PR #741, opened 2018-05-14
feature: add bootstrap grid system

### LINKED ISSUE #613: Add Grid CSS System
# User Story

### As a: 

Developer

### I want:

to use a css grid system for better & general layouting and better mobile optimization. For example the approach of Bootstrap (only grid) or [Uikit](https://getuikit.com/) is a good starting point. 

## Acceptance criteria:
- [ ] have a stable and feature-rich css grid system

## Additional info:

------

armel22:
## Description:
add bootstrap grid system

Closes #613

## Improvements
- 

## Known Issues:
_NONE_


armel22:
@PatrickSkowronek I dind f
we decided few week ago to use a grid-system and @Gargamil suggestion was either UI-kit  or boostrap Grid. So I included the bootstrap 4 Grid System to the project which is only the a css file.

Now for better responsiveness people could just use the css classes.  Do we really need an exemple for that ???? I mean it is well documented  on the official site

Gargamil:
Maybe we should mention it somewhere in the wiki, or in a file like `app.component.scss` (or somewhere else), that devs can use the grid. In addition a Link to the [relevant Bootstrap documentation](https://getbootstrap.com/docs/4.1/layout/overview/) would be very useful.

An Example does not make sense here, i think..

PatrickSkowronek:
@Gargamil @armel22 my definition of example is different. Maybe there is a usecase where you cloud already use the grid system of bootstrap? If not I will merge it.

PR #767, opened 2018-06-05
Bugfix/#753 rework translations

mleppla:
## Description:
- add missing translations in HTML-Files

## Improvements
- there should be no missing translations in HTML-Files

## Known Issues:
- there are still missing translations in snackbars/ typescript-files


mleppla:
@HPunktOchs This ticket is only for the translations in HTML files, perhaps you can create a new ticket with a 3rd iteration of the translation? In the new ticket, we can translate also the ts-files. Otherwise the issue is too big.

PR #787, opened 2018-06-19
Fixed return of dependencies a proper json object

PatrickSkowronek:
## Description:
Fix the previous version!

### DIFF
--- a/api/test/tslint.json
+++ b/api/test/tslint.json
@@ -1,0 +1,6 @@
+{
+  "extends": "../tslint.json",
+  "rules": {
+    "no-unused-expression": false
+  }
+}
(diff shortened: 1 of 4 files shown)


danielkesselberg:
Nice catch but when i apply your patch than "Content-Type" is missing from the response.

api/test/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../tslint.json",
+  "rules": {
+    "no-unused-expression": false
```
danielkesselberg: tslint complains about `expect(result).to.be.json;` as unused expression.

PR #791, opened 2018-06-21
Update validator dependency

danielkesselberg:
## Description:

Update validator dependency


danielkesselberg:
@PatrickSkowronek reverted the winston related changes with this pull request. This is only for updating the validation library.

PR #794, opened 2018-06-21
Feature/#776 dsgvo export user data

### LINKED ISSUE #776: EU-DSGVO: Export user data
# User Story

### As a: 
User

### I want:
Export all my personal Data with one click.

### so that:
A user can download all his data.

Gargamil:
## Description:
User can export his data. 

Closes #776 

## Known Issues:
- Backend-Tests are missing


api/src/controllers/ExportController.ts
```diff
@@ -89,4 +97,36 @@ export class ExportController {
...
+  async exportAllUserData(@CurrentUser() currentUser: IUser,  @Res() response: Response) {
+    // load user
+    const user = await User.findById(currentUser);
+
+    if (!user) {
+      throw new NotFoundError(`User was not found.`);
+    }
+
+    //
+    /*response.setHeader('Content-disposition', 'attachment; filename=export.json');
```
PatrickSkowronek: Why is this code comment out? Do we need it, or not.

PatrickSkowronek: ok, did you see one test failed and your Changelog entry is missing

api/src/controllers/ExportController.ts
```diff
@@ -89,4 +97,36 @@ export class ExportController {
...
     return unit.exportJSON();
   }
+
+  @Get('/user')
+  @Authorized(['student', 'teacher', 'admin'])
+  async exportAllUserData(@CurrentUser() currentUser: IUser,  @Res() response: Response) {
+    // load user
+    const user = await User.findById(currentUser);
+
+    if (!user) {
```
danielkesselberg: You could remove this if. `currentUser` is injected by express routing middleware and always available because of `@Authorized`

Gargamil: In my point of view not. Because: `IUser` is different from the `IUserMongoose`, which has more functionality. So at my point of view the User as mongoose-type is needed, because e.g. the call of `await user.exportPersonalData()` 

BUT, if i'm wrong, please show me the correct way. I would appreciate it!

danielkesselberg: Nothing wrong with `const user = await User.findById(currentUser);`. I tried to say that is technically not possible that `if (!user)` is true. `@Authorized(['student', 'teacher', 'admin'])` validates before the actual method that this is a valid request by a student, teacher or admin and pass a IUser instance.

Gargamil: oh, yes of course, you're right! Think i copied this part somewhere.. 
Thanks!

Issue #20, opened 2017-04-17
Configuration script during installation


grosspersky:
# User Story

### as: 
site builder

### i want to:
define a first admin account and other configuration data during the app installation

### as a result:
I have a secure way and easy way to define my first admin account and configuration data

### acceptance criteria:
Parameters defined during installation are used for configuration and after the installation process is finished a account with role admin exists

PR #807, opened 2018-06-26
Bugfix/#729 secure static uploads

### LINKED ISSUE #729: 🐛 BUG: Fix security leak: static upload path
# User Story

Currently the `api/uploads` can be accessed without logging into geli. https://github.com/h-da/geli/blob/dd7dbd94fc08ab901376a1bd526cbd6cd044d6d1/api/src/server.ts#L58 This should be changed. 

- [x] Done: User must be logged in.
- [x] See ¹: The user has access to the course where the file is present.
- [x] See ¹: The file is present in a unit and not just in the media manager.

¹ @torss: These two aspects have not really been achieved (see the PR for details), but are to some degree already guaranteed thanks to the secret / randomized / unguessable filenames that `geli` employs. _(Unless I am severely mistaken that is.)_

torss:
## Description:

Closes #729

This secures the `'uploads'` route, which was previously completely unprotected / public.

Due to browser API limitations it proved to be apparently impossible to modify the request headers for ``s or `<video>`s, meaning that we can't simply attach our JWT `localStorage['token']` as we do for various other API requests. And while it is possible to separately download image files via `BackendService`s `getDownload` or similar, this is work-around is not feasible for serving large videos! Thus a different approach was required:

A new, specialized `'mediaToken'` is generated in addition to the regular `'token'`. This `'mediaToken'` is only granted access to routes using the accompanying new `passportJwtMiddlewareMedia` (that means the underlying new `passportJwtStrategyMedia`). So if the `'mediaToken'` is accidentally or willfully passed on by a user, the recipient can't use it to fully impersonate the user / can't use it to gain access to the actual `geli` page functionality that requires the regular `'token'`. This is important because this new `'mediaToken'` is transmitted in form of a URL query parameter. 

This solution means that ``s and more importantly `<video>`s are now secured without compromising functionality (e.g. by always separately pre-loading entire files). There is also a beneficial side-effect to attaching the `'mediaToken'` as URL query parameter: Users can still copy the file-URLs with their attached `'mediaToken'` and send them to anyone else. Recipients of such URLs can still view the file, but only as long as the `'mediaToken'`s last. Since we apparently already employ randomized (unguessable) filenames, there is no major risk of abusing `'mediaToken'`s to gain access to files that have not already been published in some way. Nevertheless, this is not perfect and sharing is currently not really intended as a result, see `Known Issues` for more.

## Improvements
- Secured the static `'uploads'` route by means of the approach described above.
- Added a `JwtPipe` to easily append `'mediaToken'`s to file URLs.
- Added a `FileComponent` to flexibly display a single file e.g. in a new tab. It is currently used for the `Edit course` → `Media` → `Open in new Tab` functionality. (This was originally intended as an unfortunate necessity to separately download files via `BackendService`s `getDownload` - which, again, would have been bad for large videos - but it now has become a simply more flexible, Angular-controllable form to display a single file versus just opening the file-URL in a new tab.)
- Appended `'mediaToken'` to various file URLs via `JwtPipe` and adapted front-end code in general.
- Replaced hard-coded `'upload'` strings in the API with `config.uploadFolder`.
- Fixed missing `@UseBefore` middleware in `MediaController`.
- Secured `DownloadController` → `getArchivedFile` → `id` input usage.
- Fixed `picture.path` backslash issue / regression. This doesn't really have anything to do with issue #729, the problem just came up after the merge. See [this commit](https://github.com/utetrapp/geli/commit/a3bdab4c0910fdbfa1e07b856b61eca1248d4635) for details.
- Other unrelated small improvements (refactoring, fixes, unit tests), e.g. better `DownloadController` coverage via new tests with new `DeleteCache` `admin`-only API.
- Admin front-end control for `DeleteCache`, housed in a new `'Misc.'` admin sub-component.
- (Scrapped experiment of a `@Controller`-based replacement for the static `'uploads'` route: `UploadsController`. See `Known Issues` for more.)

## Known Issues:
- The URLs become quite long by attaching a `'mediaToken'`. This however should be no significant issue, since they are mostly invisible to the user (e.g. unless a user chooses to `Copy image address` or similar). Opening a new tab via Angular can route to the new `FileComponent`, which tries to gracefully embed the real requested file.
- No differentiation between different users has been achieved, so once a user knows the secret filename of a course he doesn't have access to, he can access that file with his own token indefinitely. This was meant to be solved via the custom `UploadsController` in place of the `express.static` system, but while that controller seemingly worked, it caused problems for video requests that I could not solve so far without modifying the `routing-controllers` library. Perhaps a viable future alternative would be to use `express` directly to achieve this, although this wouldn't be too elegant in comparison to everything else using the `routing-controllers`.
- Concerns have been raised in the `mpse-geli/dev` Gitter room about storing the tokens in `localStorage` due to potential XSS attacks. This is not directly related to this issue #729, but if we should switch to using cookies for authentication (which must then prevent CSRF attacks and would require pervasive changes), then we would no longer have to transmit the `'mediaToken'` as URL query parameter as well.
- I got waaay to many strong headaches from working on this issue because of all the seemingly unsolvable roadblocks I encountered here, among other things. ;)

## Possible superior solution for the future:
A better, more robust general solution for the future (independent of the mentioned potential switch to cookies above) might be to just remove the entire `'mediaToken'` system again, and instead generate random proxy-filenames for each individual file-listing request. One drawback of this would be temporary server-side storage to associate the proxy-filenames with the real paths. But perhaps this too could be mitigated by generating JWTs per file, using them directly as file ID and relying on the JWT verification for security. Then the only drawback might be computational inefficiency due to the high number of JWTs that need to be generated.


danielkesselberg:
Good work as expected ;-) I personally don't like the get parameter solution. Would it be possible to read jwt media token from local storage? I guess (correct me if i am wrong) you protected express.static with jwtMediaToken. Actual the token is read from query. I think it would be easier to read token from local storage.

torss:
@danielkesselberg You mean why the token isn't sent via the request headers, as we do for other requests? This is sadly not possible because there simply is no way to modify the headers of browser-initiated requests for ``s or `<video>`s. That's the problem I mention here in the beginning of the PR-description, and that I previously talked about in a [comment](https://github.com/utetrapp/geli/issues/729#issuecomment-398893636). In short there are only a few ways (that I know of) to transmit the token for ``s and `<video>`s:

1. Using cookies. I haven't really looked into this option, because I originally thought we (geli) explicitly choose to use `localStorage` instead of cookies to avoid the potential CSRF attacks. But now only recently in our Gitter development room @grosspersky mentioned an [article](https://dev.to/rdegges/please-stop-using-local-storage-1i04) that talks about why `localStorage` is worse (mostly due to XSS attacks) and that cookies might be better after all (since they can be configured to be unreadable from JavaScript). So this might turn out to be a solution regarding the transmission of the (media) JWT in the future after all.
2. Using the URL query parameter, as is done here. This has the (sort of) benefit that the URL can be shared, as I elaborate in the PR comment above.
3. Using a service worker to inject the authentication headers. That however has its own problems, especially w.r.t. our Angular environment. See the aforementioned [comment](https://github.com/utetrapp/geli/issues/729#issuecomment-398893636) again for some details.
4. Not using any token that is valid for multiple requests or files, see the _"Possible superior solution for the future"_-section at the end of the PR comment above.
5. _(Using some browser-extension to inject headers or handle authentication in some other way. But requiring all users to install an extension would obviously be a terrible solution for this.)_

danielkesselberg:
Thank you for your clarification. I personally would prefer the cookie over the query parameter. I think the work in the frontend becomes much more complicated by the additional query parameter.

danielkesselberg:
Its not possible to merge your branch to develop. Please merge latest develop to your branch and resolve conflicts.

torss:
@danielkesselberg Hm, doesn't look like the `develop` branch changed. Is it perhaps blocked because I requested @PatrickSkowronek as reviewer? Or because `coveralls` doesn't work currently?

torss:
@danielkesselberg Also, yes, this current solution is definitely not optimal and I hope we can achieve something better in the next semester. I should however stress that cookies alone would not solve the complete issue, since we ultimately also want to control access for specific users. E.g. block requests of users for files that are attached to a course they don't have access to. See `Known Issues` above for why a first experimental approach via a `@Controller` didn't work as easily as may be expected. Regarding ease of development in the front-end, that could also be mitigated by the _"Possible superior solution for the future"_.

Gargamil:
> Thank you for your clarification. I personally would prefer the cookie over the query parameter. I think the work in the frontend becomes much more complicated by the additional query parameter.

Thats not the best solution if you want to keep the frontend and the backend separated. For example you could use JWT tokens also in Apps and any kind of platform. Cookies are Browser-specific (yes, i know its 'only' a header, but in a portable API (used for many frontends) you really don't want to use cookies. 
@torss and me watched how facebook, twitter etc. etc. are doing this. And in fact, they use url params for this. 
We even searched for a better solution but we come to the thing with media tokens.
The solution made by @torss is at the moment the best and i think he did it well!

If someone has a better solution, please please tell us about. (Even just for own interest)

torss:
@Gargamil Do you know whether or not it would be feasible / sensible to transmit the JWTs via cookies anyway, because of the [article](https://dev.to/rdegges/please-stop-using-local-storage-1i04) @grosspersky mentioned, i.e. to mitigate XSS attacks?
_(I personally do not have the required knowledge regarding cookies to confidently assess the difficulty / potential problems that such a transition to cookies would imply for both back-end and front-end.)_

Also, what do you think of (simultaneously) implementing the _"Possible superior solution for the future"_ I proposed at the end of the initial PR-comment above? I.e. generating a JWT per requested file, so that only / exactly a single file(-path) is encapsulated by such a JWT (!), and so that only these (signed & verified) JWTs can be used to request each file? My intent there is to explicitly allow the creation of publicly sharable but expiring links, that do not have the drawback of the current approach (which is partially relying on the secret file names). The expiration time could then even be user-definable, within reason, per link.

And besides these "public" full-JWT links we should also have another, non-expiring non-JWT-containing type that is only sharable with other registered users who are allowed to access the file anyway. _(I.e. including access to the specific course that the file belongs to.)_ That could perhaps then be achieved by the current style of links (without appended tokens) and the new `FileComponent` front-end route, so that we can ensure that only authorized users have access. I think. Hmm.

Well, first I have to deal with my upcoming exams, so this will have to wait for now. ;)

torss:
@Gargamil Thanks! I agree, these issues warrant further discussion and research.
_(Which for now has to wait until after the exams, as far as I am concerned anyway. ^^)_

PatrickSkowronek:
I looked over it and is big improvement! I have a small problem with the `config.uploadFolder, We should think about to add a install script #20 . Changing the folder in the would bring some problems to the docker container and local.

torss:
@PatrickSkowronek Ah sorry, the `config.uploadFolder` [already existed](https://github.com/utetrapp/geli/pull/807/commits/b80337426fd4326721a35e21f0396dd8f14bf2df#diff-038e5ca21ac6d8192b80dcf4d2c3edc4L40) in the project (it was added [here](https://github.com/utetrapp/geli/commit/b213da0fb6dc4bf5ad1028c96b38d7a6033d0535)) and as a result I didn't think much about replacing all the hard-coded functional equivalents with the `config`-variant. ;)

PatrickSkowronek:
@torss the replace is not bad, but now there cloud be a different path than the hard coded git folder. To guarantee that it will work for others we should do that soon.

torss:
Alright, so I fixed the issue re. PR #751 I noticed today at the meeting - although I think the code from that PR should be refactored (which doesn't have anything to do with the issue #729 handled here). But now after the latest merge of the `develop` branch the `should export admin data` back-end test fails! I will be unavailable until Monday evening due to my upcoming exam, so I can't deal with this immediately.

PatrickSkowronek:
@torss you can take your time, I will look after my exams on Thursday after it.

torss:
@PatrickSkowronek Curious...when I just ran the API testsuite again (locally) I could not reproduce the test failure seen in the [previous Travis CI build](https://travis-ci.org/utetrapp/geli/builds/397791891?utm_source=github_status&utm_medium=notification). Nevertheless, I have found a small flaw in the `'should export admin data'` test and also slightly cleaned up the `ExportController` / `exportAllUserData`, which I only looked at because that's what the test relies upon. I assume that the test failed coincidentally due to the random student selection. (@Gargamil If true, then there might be a bug somewhere in the export functions used in `exportAllUserData`.)

But all of this is completely unrelated to our issue #729 here.
There are however some small-ish remaining ToDos for this PR:
* Fix a small bug I think I caused in the last merge.
* Reduce/Eliminate the reduced coverage (-0.7%) due to the merges & PR changes.

torss:
@PatrickSkowronek Ok, the coverage has been improved and I think the PR should now (once again) be ready for merging!

torss:
> @torss you can take your time, I will look after my exams on Thursday after it.

@PatrickSkowronek What's your status?

PatrickSkowronek:
@torss I'm at right now 😄

app/webFrontend/src/app/shared/pipes/jwt/jwt.pipe.ts
```diff
@@ -0,0 +1,34 @@
...
+export class JwtPipe implements PipeTransform {
+
+  private availableTypes = {
+    media: 'mediaToken',
+    regular: 'token'
+  };
+
+  transform(url: string, type: string = 'media'): string {
+    const key = this.availableTypes[type];
+    const token = localStorage[key].split(' ')[1];
```
PatrickSkowronek: This can be undefined , which is on the first login. We need to catch the undefined error!

torss: Wow, good catch! I completely missed that.

The actual problem was not the `JwtPipe` - the `localStorage[key]` there **must** never be undefined when the `transform` gets called, so it should throw an error and extra `undefined`-handling code is not needed. Instead, the real problem was that `JwtPipe` got used before login completion in the first place! This happened because of changes in one of the merges, i.e. the ones re. PR #751 with the `actualProfilePicturePath` stuff I already had somewhat annoying issues with before, as mentioned in some of the comments/commits above.
_(I think I should open a GitHub issue for refactoring that thing.)_
Parts of that (i.e. `updateProfilePicture`) get called in the `UserService`'s `setUser`, which got called in the `login` method of our `AuthenticationService` **before** the tokens were set, thus causing the `undefined` error. More precisely, the error occurs in the anonymous `AppComponent` function for `userService.data.subscribe`, and so it still occurred even without `setUser`, since that function gets called on initialization anyway, but didn't sufficiently check for empty (string) input.

And the reason for why I didn't notice this problem at all is because of another issue _(this time one for which I, and not the merged stuff, is to blame ^^)_: I simply forgot to add code that clears the `mediaToken` on logout to the `AuthenticationService`! Ha.

api/src/controllers/DownloadController.ts
```diff
@@ -229,4 +240,22 @@ export class DownloadController {
...
+   * @apiGroup Download
+   * @apiPermission admin
+   *
+   * @apiSuccess {Object} result Empty object.
+   *
+   * @apiSuccessExample {json} Success-Response:
+   *      {}
+   */
+  @Delete('/cache')
+  @Authorized(['admin'])
```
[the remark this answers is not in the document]
torss: No, and I only wrote that route because I wanted to improve the `DownloadController` tests, which I in turn only did because of the reduced coverage. _(So that was by design & doesn't really have anything to do with the PR.)_ But I concur, it would be nice to have that functionality be accessible from the admin front-end; will try to add that now.

app/webFrontend/src/app/shared/components/responsive-image/responsive-image.component.html
```diff
@@ -1,6 +1,5 @@
 <picture class="responsive-image">
   <source *ngFor="let breakpoint of this.responsiveImageData.breakpoints" media="(min-width: {{breakpoint.screenSize}}px)"
-          srcset="api/{{breakpoint.pathToImage}}{{breakpoint.pathToRetinaImage ? ' 1x, ' + breakpoint.pathToRetinaImage + ' 2x' : ''}}">
+          srcset="api/{{breakpoint.pathToImage | jwt}}{{breakpoint.pathToRetinaImage ? ' 1x, api/' + (breakpoint.pathToRetinaImage | jwt) + ' 2x' : ''}}">
```
PatrickSkowronek: This route is not working properly! The old route does still work. I added a new picture and it does not work anymore.

torss: Thanks, I think this must have been broken by [this](https://github.com/utetrapp/geli/pull/807/commits/b6e5d472afa418189ddfc0f10340068f489c91b1).
Will fix it ASAP - After Sleeping Again :P

torss: Never mind, the [commit](https://github.com/utetrapp/geli/pull/807/commits/b6e5d472afa418189ddfc0f10340068f489c91b1) I previously referred to wasn't the issue - it did in fact provide the solution! The actual problem discovered here is that the `ResponsiveImageService` didn't account for absolute directory paths. _(Which also is what was fixed in said commit, but for user profile pictures.)_
Should be fixed now!

PR #824, opened 2018-08-21
Make e2e login test more stable

danielkesselberg:
Follow-Up #823


danielkesselberg:
@PatrickSkowronek I think it's better to deactivate this test for now so that nobody gets blocked by it.

PatrickSkowronek:
Ok, @danielkesselberg still cool that the rest works 😄

PR #829, opened 2018-09-12
Fixed config for docker build with new forum

PatrickSkowronek:
fixed the http 405 error

## Description:

Currently the new Forum send data back with a POST. Nginx will refuse this with a 405.

## Improvements
- A workaround for it

## Known Issues:
_NONE_


PatrickSkowronek:
@danielkesselberg thx mate, this is only a fix. In the future we need to see if we can change socketIO to fix this problem

PR #849, opened 2018-10-12
Feature/#840 JWT cookie authentication

### LINKED ISSUE #840: 🍪 Switch to JWT-cookie-based authentication
# User Story

### As a:

developer

### I want:

to have a `HttpOnly` JWT-cookie-based authentication for all current non-public requests instead of the separate URL-attached `mediaToken`s for media (e.g. image) requests as implemented by #807 (with tokens being stored in `localStorage`)

### so that:

the authentication system - and front-end development process, which currently has a special `JwtPipe` helper to attach `mediaToken`s - can be simplified and protection against XSS attacks is increased.

## Acceptance criteria:
- [x] A `HttpOnly` cookie with the primary JWT token is set.
- [x] All authentication now uses this cookie('s JWT token).
- [x] On log-out the cookie is cleared.
- [x] Old `localStorage`-based token infrastructure is removed, including the `mediaToken` and the `JwtPipe`.

## Additional info:

I'm not 1000% certain that there aren't any special cookie-related issues that could arise, so please alert me immediately if you know better!

torss:
## Description

Closes #840

Switches the entire authentication system to use a single JWT stored and transmitted via a cookie.
The cookie is `HttpOnly` and has `SameSite` set to `'Strict'` to protect against CSRF attacks.

Previously we had two tokens: A primary `'token'` and a `'mediaToken'`, both stored in `localStorage`. The `'mediaToken'` was a relatively recent addition in #807 to secure the static `'uploads'` route for images and such. In contrast to a cookie this `'mediaToken'` had to be sent via URL query parameter, see the #807 PR for details.

## Improvements

* Reduced XSS attack surface: `localStorage` tokens could (only) be read via JavaScript. Now the new `HttpOnly` cookie token can't be accessed via JavaScript at all, it is (re)set via HTTP requests.
* General simplification of the authentication system: The cookie is sent automatically by the browser, while the `localStorage` tokens had to be handled via JavaScript.
* Removal of the `'mediaToken'` system: Since the `'mediaToken'` system was only implemented due to browser API limitations #807 it was now possible to removed it completely, including the `'JwtPipe'`.
* Stricter `'uploads'` route securiy: A user now has to be logged-in to access files of the `'uploads'` route, making it impossible to accidentially share/leak a URL with attached `'mediaToken'`.

## Known Issues

* The cookie does not have the `Secure` attribute (i.e. HTTPS-only) set - maybe this should be configurable to implicitly prevent accidental non-HTTP hosting.
* The stricter `'uploads'` route security also implies that it is now impossible for users to quickly share a non-public file URL with people who don't have an account / aren't logged-in. This should probably be solved by providing proper, explicit public URL creation facilities. _(This isn't really a new issue because it was only possible to share such URLs due to the lax / non-existent security.)_
* Apparently the chat system is currently broken _(independent of this PR)_, so I couldn't test whether the switch to cookie authentication works.
* _I also haven't been able to fully conduct manual testing for file access because the course media editor seems to be broken right now; possibly related to the broken chat system?_


torss:
While this should be complete, please don't merge this to `develop` quite yet, because we still have to complete release 0.8 first and this PR basically overwrites its predecessor from that time!

danielkesselberg:
@torss i merged latest develop and resolved merge conflicts. Could you have another look at `@types/cookie` dependency? I think (https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies) it should be save to place it in devDependencies but i'm note sure :smile:

torss:
@danielkesselberg Probably yes! At least iirc I had no specific reason not to install that under `devDependencies`. Could the same thing apply to `@types/sharp` and `@types/socket.io` (the only other two `@types/[...]` installed under `dependencies`)?

_(Also a sidenote re. the `cookie` package: I've only used it in `passportJwtStrategyFactory.ts` and it thus shouldn't be hard to remove it completely as a dependency, if desired.)_

danielkesselberg:
I have no objections to `cookie` as dependency :+1:

torss:
@danielkesselberg Well, I don't know if I would call it "a lot" of research, but hopefully it was enough. xD
Regarding the `@types/[...]` in `dependencies` vs. `devDependencies` I see you have already changed / fixed that for `@types/cookie`, but - to reiterate my question above - what about `@types/sharp` and `@types/socket.io` under `dependencies`? Does someone know whether they too can be moved to `devDependencies`? _(I assume so, but maybe I'm missing something.)_

danielkesselberg:
https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

I guess `@types/sharp` and `@types/socket.io` belongs to `devDependencies` as well. But this is another story.

PR #854, opened 2018-10-17
Exclude dependabot from changelog.md check

danielkesselberg:
## Description:

We decided to use https://dependabot.com for dependency updates. dependabot is not able to add a new line to changelog.md. Its required to exclude pull requests from dependabot from changelog.md check.

## Improvements
- Early exit when not a pull request
- https://www.shellcheck.net/ is happy

## Known Issues:
_NONE_


aeimer:
@danielkesselberg should work. I have some little adaption I would like to see fixed. Give me time until tomorrow

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
 
 # Path to this file
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# Path the script was called from
-IPWD="$(pwd)"
```
PatrickSkowronek: We have a convention on the shell files. All files define the pwd.

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
...
-# Import shared vars
-. ${DIR}/_shared-vars.sh
 
-echo
-echo "+++ Check if changelog was updated +++"
```
PatrickSkowronek: We use echo the show what will executed, even when there will nothing to be executed.

PatrickSkowronek:
We have a convention in the shell files, which @danielkesselberg properly not really know? Please add it. The rest looks good, but I would test it locally.

.travis/changelog.sh
danielkesselberg: Not required here.

.travis/changelog.sh
danielkesselberg: https://github.com/geli-lms/geli/blob/60d2e6a8f6304fefd3fb5adcc18d2d4a67ce07ca/.travis/changelog.sh#L54-L55

Default case when a pull request. Every other case is covered.

.travis/changelog.sh
aeimer: Definetly on @PatrickSkowronek site! All shell files should start the same way, so we have everywhere the same "environment"

.travis/changelog.sh
aeimer: I'm on @PatrickSkowronek side, we always print the task we do on the first line. We should with this tradition. Makes sense in the travis log

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
...
-echo
+# shellcheck source=_shared-vars.sh
```
aeimer: Please remove this comment

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
...
+. "${DIR}/_shared-vars.sh"
 
+if [[ ${TRAVIS_PULL_REQUEST} == false ]]; then
+  echo -e "${GREEN}+ Skip changelog check (no pull request)"
```
aeimer: Always end with ${NC} if you define a color before

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
...
+  echo -e "${GREEN}+ Skip changelog check (dependency update)"
+  exit 0;
+fi
 
-if [ "$TRAVIS_PULL_REQUEST" != "false" ] && [ "$TRAVIS_BRANCH" == "develop" ]; then
-  echo "+ detected pull request from ($TRAVIS_PULL_REQUEST_BRANCH) to $TRAVIS_BRANCH"
-  curl --silent https://raw.githubusercontent.com/geli-lms/geli/$TRAVIS_BRANCH/CHANGELOG.md \
+if [[ ${TRAVIS_BRANCH} == "develop" ]]; then
+  echo "+ Pull Request from ${TRAVIS_PULL_REQUEST_BRANCH} to ${TRAVIS_BRANCH}"
+  curl --silent https://raw.githubusercontent.com/geli-lms/geli/"${TRAVIS_BRANCH}"/CHANGELOG.md \
```
aeimer: if you want to use " do it with the whole param: `curl --silent "https://raw.githubusercontent.com/geli-lms/geli/${TRAVIS_BRANCH}/CHANGELOG.md"`

.travis/changelog.sh
```diff
@@ -2,43 +2,54 @@
...
+if [[ ${TRAVIS_BRANCH} == "master" ]]; then
+  echo "+ Pull Request from ${TRAVIS_PULL_REQUEST_BRANCH} to ${TRAVIS_BRANCH}"
+  curl --silent https://raw.githubusercontent.com/geli-lms/geli/"${TRAVIS_BRANCH}"/CHANGELOG.md \
   | diff CHANGELOG.md - \
   | grep -P '^< ## \[\[\d{1,3}\.\d{1,3}\.\d{1,3}\].*\] - \d{4}-\d{2}-\d{2} - .{10,}' - -q
+  RESULT=$?;
 
-  if [[ $? == 0 ]]; then
-    echo -e "${GREEN}+ Update in CHANGELOG.md found, exit${NC}"
+  if [[ ${RESULT} -eq 0 ]]; then
```
aeimer: Why do you use the `-eq`?

.travis/changelog.sh
danielkesselberg: Annotation required by shellcheck

.travis/changelog.sh
danielkesselberg: Its common to use -eq to compare numbers

.travis/changelog.sh
aeimer: But it's not needed here. It's good that you check your code, but please remove it here. It just confuses...

danielkesselberg:
@PatrickSkowronek @aeimer apart from your suggestions, you're seeing some problems with the functionality?

.travis/changelog.sh
PatrickSkowronek: Do you want to add this as a check? Otherwise can you explain why we need this.

danielkesselberg: I would appreciate it but I'm not sure if I have enough time.

aeimer: @danielkesselberg I don't understand what you want to say...

aeimer:
@danielkesselberg yeah should work, you propably need to merge and try it...

danielkesselberg:
image

PR #855, opened 2018-10-17
Rebuild phantomjs inside docker

danielkesselberg:
## Description:

We copy `node_modules` into the docker image. I suppose that the phantomjs version for travis ci does not work for the docker image. This pull request add package.json / package-lock.json into the container and execute npm rebuild during docker build.
 https://github.com/geli-lms/geli/blob/fdaec950f59f628a8cdb9ac5813da3e01d652c2b/.docker/api/Dockerfile#L19

> PhantomJS needs to be compiled separately for each platform. This installer finds a prebuilt binary for your operating system, and downloads it.

> If you check your dependencies into git, and work on a cross-platform team, then you need to tell NPM to rebuild any platform-specific dependencies. Run

https://www.npmjs.com/package/phantomjs-prebuilt#cross-platform-repositories

> There is no requirement to install Qt, WebKit, or any other libraries. It however still relies on Fontconfig (the package fontconfig or libfontconfig, depending on the distribution).

https://www.npmjs.com/package/phantomjs-prebuilt#linux-note

## Improvements
- A working version ofr phantomjs is shipped with docker image
- Export as PDF work

## Known Issues:
_NONE_

### DIFF
--- a/.docker/api/Dockerfile
+++ b/.docker/api/Dockerfile
@@ -9,6 +9,8 @@ EXPOSE 80
 
 CMD node server.js
 
+RUN apt-get update -y && apt-get install libfontconfig1 -y
+
 RUN mkdir -p /usr/src/app/uploads
 RUN mkdir -p /usr/src/app/uploads/courses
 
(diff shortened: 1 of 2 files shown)


aeimer:
If it works as suspect perfect. Looks good

Gargamil:
Just tested by locally building the api container. It works!

PR #856, opened 2018-10-17
Prepare changelog for bugfix & tweak release 0.8

danielkesselberg:
## Description:

Close https://github.com/geli-lms/geli/projects/11


danielkesselberg:
> @danielkesselberg the build failed, add a Next section and it should be good

No. The check https://github.com/geli-lms/geli/blob/097c26f68665e5a96c46b2c5df8ed5ae3df2dc2b/.travis/changelog.sh#L19 is looking for a line starting with ` - at least 8 characters`

PatrickSkowronek:
wired it should actually work.

PR #869, opened 2018-10-18
Bugfix/#857 course chat rooms broken

### LINKED ISSUE #857: 🐛 BUG: File upload is broken on staging
# User Story
File Upload / Mediamanager is broken totally on staging. 
grafik

File upload is broken . 

## Acceptance criteria:
- [ ] fix Mediamanager / file upload

torss:
## Description:

Closes #857

## Improvements:
- Fixes broken `Course` `chatRooms` id extraction re. (at least) the `GetCourseEdit` (`/api/courses/:id[...]/edit`), i.e. `Course.toObject`, that broke (at least) "Edit course - Media" by providing invalid information back to the `PutCourse` route.
- Fixes missing `Course` constructor `chatRooms` assignment in the front-end. _(I am not sure whether this bug actually caused any noticeable issues or not.)_

## Known Issues:
_NONE_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ cache:
     - app/webFrontend/node_modules
     - .travis/node_modules
 
+before_install:
+  # check for change in changelog
+  - .travis/changelog.sh
+
 before_script:
   # required for chrome to start properly
   - export DISPLAY=:99.0
@@ -62,8 +66,6 @@ install:
   - cd ..
 
 script:
-  # check for change in changelog
-  - .travis/changelog.sh
   # check for versions
   - .travis/check-version.sh
   # run license checker and write license.json / dependencies.ts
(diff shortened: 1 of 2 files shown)


api/src/models/Course.ts
```diff
@@ -121,11 +121,7 @@ const courseSchema = new mongoose.Schema({
           }
         }
 
-        if (doc.populated('chatRooms') !== undefined) {
-          ret.chatRooms = ret.chatRooms.map((chatRoom: any) => {
-            return chatRoom.toString();
-          });
-        }
+        ret.chatRooms = ret.chatRooms.map(extractMongoId);
```
torss: This probably broke something according to the unit tests, check it again.

torss: Alright, looks like the issue was indeed caused by this not checking for `chatRooms` being defined. The problem manifested in the `"should export student data"` / `"should export teacher data"` unit tests, because they request the `exportAllUserData` route, which uses the `Course` `exportPersonalData` function, which in turn calls `Course.find(conditions, 'name description -_id')` (i.e. excluding `chatRooms` due to the given `mongoose` "projection"-parameter argument).

torss: There are also some things that could be refactored / fixed, e.g. maybe the `'should export student data'` test and the `Course` `exportJSON` method, but that can wait and isn't relevant here.

PR #893, opened 2018-10-20
Fix class definition for main content area

grosspersky:
## Description:

This fixes an issue with the maximum width of the main content area not being set anymore. Because of this the main content area would use more space than the header.

With this PR the original behaviour is restored, while keeping the padding adjustments for file displays.

## Improvements
- Maximum width of main content area is working again

## Known Issues:
_NONE_


danielkesselberg:
live0 7
staging0 8

PR #879, opened 2018-10-18
Prepare changelog for bugfix & tweak release 0.8

danielkesselberg:
## Description:

Close https://github.com/geli-lms/geli/projects/11

PR #907, opened 2018-10-23
Add fileUnitType for FileUnit

danielkesselberg:
## Description:

For some units the field fileUnitType is missing. Save the record
again does the trick.

## Improvements
- No warning when migration data

## Known Issues:
_NONE_


danielkesselberg:
Add `20181019-fileUnit.ts` to #879 after merge

PR #919, opened 2018-10-25
PDF export: Use random hash as file name for temp file

### LINKED ISSUE #913: PDF export fails randomly
# User Story

### As a: 
user of the GELI platform

### I want:
to export the course as PDF

## Acceptance criteria:
- [x] Display download button again
- [x] Rename temp file to be unique
- [x] Fix pdf generation and download
- [ ] Use `populate` (if possible) to remove `localLecture` and `localUnit`
- [ ] Invalidate cache when course / lecture / unit is edited

dboschm:
## Description:
Pdf export will no longer use the same temp.pdf file for all users (parallel downloading not supported).
Instead the temp file has a hash in its name now. (the hash is calulated by current Date and Userid)

Ref #913 

## Improvements
- Simultaneous Pdf Download

## Known Issues:
_NONE_

### DIFF
--- a/api/src/controllers/DownloadController.ts
+++ b/api/src/controllers/DownloadController.ts
@@ -26,8 +26,6 @@ const pdf =  require('html-pdf');
 const phantomjs = require('phantomjs-prebuilt');
 const binPath = phantomjs.path;
 
-const PDFtempPath = config.tmpFileCacheFolder + 'temp.pdf';
-
 // Set all routes which should use json to json, the standard is blob streaming data
 @Controller('/download')
 @UseBefore(passportJwtMiddleware)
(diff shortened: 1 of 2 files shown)


danielkesselberg:
@dboschm please have look :smile: 

- extracted the hash generation into a custom method.
- create temp file name inside loop for lectures when individual pdf.

api/src/controllers/DownloadController.ts
```diff
@@ -204,6 +202,9 @@ export class DownloadController {
         let lecCounter = 1;
         for (const lec of data.lectures) {
 
+          // create hashed pdf file
+          const tempPdfFileName = this.tempPdfFileName(user);
```
lukas-schardt: @dboschm My bad, i know i wanted it inside the loop, but i just realised we could put it outside to save a little bit of resources, as we are removing the file after every write. But not really a major issue in my opinion

PR #967, opened 2018-11-01
Fix delete lecture api endpoint type mismatch

### LINKED ISSUE #965: Fix Lecture Controller Types
There is a compiler Error in Lecture Controller
grafik

dboschm:
## Description:
Refactoring of the deleteLecture Endpoint 

Closes #965 

## Improvements
- fixed typemismatch in deleteLecture
- use async await
- Authorization Check if user can change Course and delete unit

### DIFF
--- a/api/src/controllers/LectureController.ts
+++ b/api/src/controllers/LectureController.ts
@@ -112,6 +112,8 @@ export class LectureController {
    * @api {delete} /api/lecture/:id Delete lecture
    * @apiName DeleteLecture
    * @apiGroup Lecture
+   * @apiPermission teacher
+   * @apiPermission admin
    *
    * @apiParam {String} id Lecture ID.
    *
@@ -125,15 +127,17 @@ export class LectureController {
   @Authorized(['teacher', 'admin'])
   @Delete('/:id')
   async deleteLecture(@Param('id') id: string, @CurrentUser() currentUser: IUser) {
-    const course = await Course.findOne({where: {lectures: id}});
+    const course = await Course.findOne({ lectures: id});
     if (!course) {
       throw new NotFoundError();
     }
-    if (!course.checkPrivileges(currentUser).userCanEditCourse()) {
+    if (!course.checkPrivileges(currentUser).userCanEditCourse) {
       throw new ForbiddenError();
     }
-    const success = await course.update({}, {$pull: {lectures: id}});
+    await course.update({}, {$pull: {lectures: id}});
 
-    return Lecture.findByIdAndRemove(id) && success;
+    await Lecture.findByIdAndRemove(id);
+
+    return {result: true};
   }
 }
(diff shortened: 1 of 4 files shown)


api/src/controllers/LectureController.ts
```diff
@@ -120,10 +122,18 @@ export class LectureController {
...
+    const course = await Course.findOne({where: {lectures: id}});
+    if (!course) {
+      throw new NotFoundError();
+    }
+    if (!course.checkPrivileges(currentUser).userCanEditCourse()) {
+      throw new ForbiddenError();
+    }
+    const success = await course.update({}, {$pull: {lectures: id}});
+
+    return Lecture.findByIdAndRemove(id) && success;
```
danielkesselberg: Could you check that success is an object you can use in a boolean operation? course.update returns an https://docs.mongodb.com/manual/reference/method/db.collection.update/#writeresults-update object.

danielkesselberg: When i understand https://mongoosejs.com/docs/api.html#model_Model.findByIdAndRemove correctly than we return `Query && WriterResults`?

dboschm: You are Right. Mongoose seems to be more complex than i thougt. I'll check/fix/whatever this.

danielkesselberg: Good luck! Maybe @grosspersky can help :+1:

dboschm: What do you think about this solution? We do it in a similar way in the [AuthController](https://github.com/geli-lms/geli/blob/58e2a68356405acc6d79a288c3ccb34056beb96b/api/src/controllers/AuthController.ts#L151)

torss: Let's maybe just return an empty object on success?

danielkesselberg: https://github.com/Automattic/mongoose/issues/2647 I think you have to pass either a dummy function as callback or call .exec on the returned query object to remove the lecture.

dboschm: @torss 1. I return `{result: true}` because this is what the ApiDoc says of what should be returned. May be response an empty Object will work, too. 
2. I Don't like appending thens to thens, especially if Idon't Know wheter this is a Promise.then() or a Query.then(). So I took async/await like in the other controllers.

torss: @dboschm 1. Yeah, I think the response / ApiDoc can be changed because the only request in the front-end seems to be [here](https://github.com/geli-lms/geli/blob/bd35bb1a1ff71f5a3db65921e62f9cc4056db660/app/webFrontend/src/app/course/course-edit/general-tab/general-tab.component.ts#L224), which discards the result.
2. Yes, it should use async/await. I think @danielkesselberg was talking about how the original `return Lecture.findByIdAndRemove(id) && success;` wouldn't have executed without `.exec()` or a dummy function.

PR #984, opened 2018-11-05
Bugfix/#877 notifications for invisible courses

### LINKED ISSUE #877: Bug: Notifications of Invisible Courses / Units
A Student gets Course information of invisible courses and units via notification

## Acceptance criteria:
- [ ] filter notifications by visibility of courses and units 

## Estimate:
2h

MichaelNarkus:
## Description:
Notifications will no longer be created for invisible lectures and inactive courses

Closes #877

## Improvements
- 

## Known Issues:
_NONE_


api/src/controllers/NotificationController.ts
```diff
@@ -92,6 +92,12 @@ export class NotificationController {
   }
 
   async createNotification(user: IUser, text: string, changedCourse?: ICourse, changedLecture?: ILecture, changedUnit?: IUnit) {
+    if (changedCourse && !changedCourse.active) {
+      return;
+    }
+    if (changedUnit && !changedUnit.visible && changedCourse.active) {
```
lukas-schardt: Could be simplified to `if (changedUnit && !changedUnit.visible)`. Active will never be `false` at this point because of the preceding if-block.

dboschm: must be changed to `if (changedUnit && !changedUnit.visible)`. because changedCourse.active will throw an "can not read property active of undefined error" If a teacher changes a visible unit.

dboschm: The reason why `if (changedUnit && !changedUnit.visible)` failed, was in the test case. Instead of sending a populated unit object, only the unit id was sent in the request and !unitId.visible always resolves to true. About 4h work to find such a stupid bug.

PR #1002, opened 2018-11-08
Bugfix/#989 chat message forgery

### LINKED ISSUE #989: 🐛 BUG: Chat message forgery 🔥
The `ChatServer` currently accepts arbitrary `ISocketIOMessage`s _(which consist out of one `IMessage` and one `ISocketIOMessageMeta`)_. This means that it is possible for any logged-in user _(connected to the `ChatServer`)_ to forge malicious chat messages in various ways:

* The perhaps most important one is that anyone can be perfectly impersonated once an attacker has received a single chat message sample _(since the messages seen in the front-end even contain the `author` ID, besides - of course - the display name `chatName`)_.
* Furthermore the destination _(`room` and `parent`)_ can be forged arbitrarily as well, thus exacerbating the impersonation attack vector by allowing an attacker to post to chat rooms of supposedly non-accessible courses.

Since the attack is also basically untraceable _(except for potential side channels, e.g. login time window discrepancies)_ the potential for abuse is rather large.

Addendum: Seems like the `author` IDs stored in the messages _(and thus sent to the clients)_ are the actual user IDs, annihilating the anonymity of the forum!

## Acceptance criteria:
- [x] Take only the absolutely necessary bare minimum of information from clients _(i.e. only the JWT authorization info, the actual chat message and the destination)_ to fix this flaw and tighten chat security in general.
- [x] Also check that the user is authorized to post to the designated destination.
- [x] Addendum: Ensure that the forum is actually anonymous. E.g. by just not sending the `author` IDs to the clients, if possible _(if not, more changes are required)_.
- [x] Maybe refactor the code, e.g. share interface files? _(Refactoring seems to be indirectly necessary anyway to properly fix the security issues.)_
- [x] Ensure that the user is authorized to connect to the requested chat room.
- [x] Fix the related `MessageController` security issues.
- [x] Fix any non-deterministic unit tests that now fail randomly.

## Additional info:
This security issue was discovered during the recent #853 audits:

* [`ChatServer`](https://github.com/geli-lms/geli/issues/853#issuecomment-435975501) 
* [`MessageController`](https://github.com/geli-lms/geli/issues/853#issuecomment-435192879)

torss:
## Description:

Closes #989

Deals with security issues of the chat _(or "forum")_ system, that allowed even an attacker with only a `student`-role to easily and untraceably impersonate any user _(only needing the user ID and "chat name", which were both contained in any chat message)_. The inclusion of the user ID _(via the `author` field)_ also meant that the anonymity of the chat system was actually broken as well. In addition, the flaws allowed any logged-in user to connect and post to any chat room _(e.g. even when the course isn't accessible)_, although this was likely less problematic since the chat room ID still needed to be known.

## Improvements:
- Chat impersonation vulnerability fixed.
- Broken chat anonymity fixed.
- Incomplete chat room authorization vulnerability fixed.
- Missing chat parent message authorization vulnerability fixed.
- Removed 2 vulnerable and unused `MessageController` routes.
- Nondeterministic `/api/message` & `/api/chatRoom` unit test authorization bugs fixed.
- Added `/api/message` access denial unit tests.
- Refactoring related to the chat system.
- Some unrelated opportunistic refactoring; including a new `TestHelper` class for shared unit test functionality and the new `extractSingleMongoId` variant of `ExtractMongoId`.

## Known Issues:
_NONE_


torss:
Well, looks like the associated unit test(s) weren't designed with authentication in mind and now non-deterministically fail _(i.e. contain random student/course selection)_.

api/fixtures/FixtureUtils.ts
```diff
@@ -150,6 +150,13 @@ export class FixtureUtils {
     return Unit.findById(unitId);
   }
 
+  public static async getSimpleChatRoomSetup(hash?: string) {
+    const course = await FixtureUtils.getRandomCourse() as ICourseModel;
+    const room = await FixtureUtils.getRandom<IChatRoom>(course.chatRooms, hash);
+    const roomId = room._id.toString();
```
danielkesselberg: I would prefer `extractMongoId` here

api/test/controllers/ChatRoom.ts
```diff
@@ -1,55 +1,29 @@
...
       expect(result).to.have.status(404);
     });
 
     it('should return chat room', async () => {
-      const student = await FixtureUtils.getRandomStudent();
-      const course = await FixtureUtils.getRandomCourse() as ICourseModel;
-
-      const room = getRandomRoomFromCourse(course);
-      const roomId = room._id.toString();
+      const {roomId} = await FixtureUtils.getSimpleChatRoomSetup();
```
danielkesselberg: Looks weird :rofl: Does `const roomId` work as well?

api/fixtures/FixtureUtils.ts
torss: Hmm, I adopted that `toString` variant from your original `/api/message` unit test code (`Message.ts`), because I thought there should be no real need for `extractMongoId` here _(since the content of `course`/`room` shouldn't change unless the `getRandomCourse` is modified to populate `room` or something)_. But now I think to remember someone (you?) mentioning that an upcoming `mongoose` updates might break such `toString` code, is that correct?

Anyway, yeah, perhaps `extractMongoId` does make slightly more sense for the sake of clarity regardless. And I also forgot to remove that superfluous `as ICourseModel`...

torss: Done! I've added a new `extractSingleMongoId` function though, because I'm not really happy with how my original `extractMongoId` apparently affects the TypeScript information. I.e. that utility still has room for improvement I think.

api/test/controllers/ChatRoom.ts
torss: What do you mean? 😮

danielkesselberg: `const {roomId}` vs `const roomId`

torss: Do you mean replacing `const {roomId}` directly with `const roomId`? That's not possible because `await FixtureUtils.getSimpleChatRoomSetup()` returns `{course, roomId}` to enable a more flexible usage _(see `test/controllers/Message.ts` where the `course` is needed for `testAccessDenial`)_.

api/fixtures/FixtureUtils.ts
```diff
@@ -150,6 +151,13 @@ export class FixtureUtils {
     return Unit.findById(unitId);
   }
 
+  public static async getSimpleChatRoomSetup(hash?: string) {
+    const course = await FixtureUtils.getRandomCourse();
+    const room = await FixtureUtils.getRandom<IChatRoom>(course.chatRooms, hash);
+    const roomId = extractSingleMongoId(room);
+    return {course, roomId};
```
danielkesselberg: Why do you return a course object and a room id? I think we should return the room. What do you think? When you have time you could introduce a small doc block? What is returned when method called without hash?

api/fixtures/FixtureUtils.ts
danielkesselberg: I like your `extractSingleMongoId` and `extractMongoId` way over `x._id.toString()` because its much easier to understand. We need this kind of hack because of the weird database model design. By default a virtual getter for id is included (https://mongoosejs.com/docs/guide.html#id) but not available when `ICourse` or `IWhateverInterface` is used because the interfaces does not extend from `mongoose.Schema`.  

The approach is based on https://github.com/Appsilon/styleguide/wiki/mongoose-typescript-models and maybe we can improve this by using `IWhateverModel` because the virtual id getter should be available there.

api/fixtures/FixtureUtils.ts
torss: As commented above, there is a reason why it also returns the course _(i.e. that's required in the `testAccessDenial` in `test/controllers/Message.ts`)_.

The `hash` parameter is simply passed to `getRandom` which apparently uses the "`hash`" as a RNG seed. I currently don't know if that functionality is even used; I only added the parameter to `getSimpleChatRoomSetup` because various other `FixtureUtils` functions do the same. So we could remove that there and maybe even remove it from `getRandom` etc. altogether if it's unused _(and/or refactor `FixtureUtils` for easier handling of static seeds, e.g. by going for non-static functions and a constructor that takes the seed or something)_. But that's beyond the scope (and purpose) of this issue I reckon.

danielkesselberg: To clarify my first question. Why do you return course and roomId? I would prefer course and room.

api/fixtures/FixtureUtils.ts
torss: Ah, ok!
Yes I think you are right, the `extractMongoId` functions are more self-explanatory than `toString`.

api/fixtures/FixtureUtils.ts
torss: Ah I see. I do that because currently `getSimpleChatRoomSetup` is a rather specific function, i.e. currently only for the two chat unit test files, and returning the `room` instead of the `roomId` wouldn't help there. If some code in the future requires the original `room` as well, it can simply be added to the object returned by `getSimpleChatRoomSetup` without having to modify the current use cases.

PR #1018, opened 2018-11-15
UglifyJs => Terser

danielkesselberg:
## Description:

https://github.com/angular/angular-cli/issues/5775#issuecomment-433013783

Build could be faster when terser (https://www.npmjs.com/package/terser) is used instead of uglify-js. 

**Uglify-Js**

image

**Terser**
image

## Improvements
- Faster build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Update mongoose to 5.2.x. [#1004](https://github.com/geli-lms/geli/pull/1004)
 - Update contributors list. [#1007](https://github.com/geli-lms/geli/issues/1007)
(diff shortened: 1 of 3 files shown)


danielkesselberg:
**Uglify-Js**

image

**Terser**
image

danielkesselberg:
Angular 7 is not on the roadmap (yet).

PatrickSkowronek:
@danielkesselberg the upgrade should be not that problematic, like the update from 5->6

danielkesselberg:
How is this related to this pr? Yes `teser` is default for angular 7 but I try to reduce the build time a bit and not upgrade angular. We will discuss the angular upgrade at some point but there is no need to rush. Angular 6 is supported until Nov 2019.

danielkesselberg:
Just for the record. Build is not faster on travis with terser. 

**Uglify-Js**

image

**Terser**

image

Issue #1003, opened 2018-11-08
Update Mongoose dependency


dboschm:
# User Story

### As a: 
Developer

### I want:
to use the latest version of mongoose

### so that:
i'm up to date

## Acceptance criteria:
- [ ] Bump Mongoose 
- [ ] Bump Mongoose Types 
- [ ] Fix all bugs and deprecation warnings 😵

## Additional info:
dependabot has already 2 PRs on this (#991 #985), but you should update the types and the mongoose framework simultaneously. 

##Estimate:
15 h

PR #1027, opened 2018-11-21
Replace update with updateOne

danielkesselberg:
## Description:

Ref #1003 

> Like remove(), the update() function is deprecated in favor of the more explicit updateOne(), updateMany(), and replaceOne() functions. You should replace update() with updateOne(), unless you use the multi or overwrite options.

https://mongoosejs.com/docs/deprecations.html#-update-


danielkesselberg:
@dboschm Do you think we could replace the update call like below? https://github.com/geli-lms/geli/blob/a3a16f50d74145973345b6ce920a703d9f305b13/api/src/controllers/LectureController.ts#L132

We are removing the lecture from all courses right?
```
await Course.updateMany({}, {$pull: {lectures: id}}); 
```

api/src/controllers/CourseController.ts
```diff
@@ -752,10 +752,7 @@ export class CourseController {
 
     await image.save();
 
-    await Course.update({ _id: id }, {
-      image: image._id
-    });
-
+    await Course.updateOne({_id: id}, {image: image._id});
```
danielkesselberg: @dboschm @torss i think we should use `$set` here? https://docs.mongodb.com/manual/reference/method/db.collection.update/#update-parameter

torss: Hm I've taken a quick look at your link and at <https://mongoosejs.com/docs/api.html> re. `$set` and I'm not sure how this works, but I think an extra `$set` might not be needed here, since the single `image: image._id` could be automatically understood as `$set` operation. _(BTW there also is this [Query.prototype.set()](https://mongoosejs.com/docs/api.html#query_Query-set) method, which does the same thing as a `$set`.)_

danielkesselberg: I'm not sure about that either.

> If the <update> document contains update operator modifiers, such as those using the $set modifier, then:
> 
>     The <update> document must contain only update operator expressions.
>     The update() method updates only the corresponding fields in the document.

We are not using $set here so the above is not true?

> If the <update> document contains only field:value expressions, then:
> 
>     The update() method replaces the matching document with the <update> document. The update() method does not replace the _id value. For an example, see Replace All Fields.
>     update() cannot update multiple documents.

`{image: image._id}` so this is true? But then the whole document should be replaced? I wonder why this works actually. After adding a picture the course should be broken :sunglasses:

torss: I agree that
```
course.image = image;
await course.save();
```
looks cleaner, but might it also be less efficient than the `updateOne` approach?
_(Not that we care too much about efficiency, so I guess it doesn't really matter. ^^)_

torss: Also, if we have `course`, couldn't we write `course.updateOne({image})`? _(See <https://mongoosejs.com/docs/api.html#document_Document-updateOne>.)_

danielkesselberg:
@torss @dboschm please review again ;-)

api/src/controllers/CourseController.ts
danielkesselberg: > TypeError: course.updateOne is not a function

Looks like updateOne is not available for 5.2.x

torss: Ha, so much for that. xD

danielkesselberg: `course.save` is using `update` under the hood: https://github.com/Automattic/mongoose/commit/3a28593850890479b4e04c4df757df3dc58bdae6

PR #1031, opened 2018-11-24
Enable vendor chunk / Disable progress

danielkesselberg:
## Description:

Enable vendor chunk: Move 3rdparty dependencies into one file.
Disable progress: Hide progress output while building


app/webFrontend/angular.json
```diff
@@ -16,6 +16,7 @@
             "main": "src/main.ts",
             "tsConfig": "src/tsconfig.app.json",
             "polyfills": "src/polyfills.ts",
+            "progress": false,
```
dboschm: @danielkesselberg What does `progress: false` do?

danielkesselberg: No output during build

PR #1066, opened 2018-12-11
Secure course picture upload/remove

### LINKED ISSUE #1053: Course list broken when course image in invalid state
## Steps to reproduce

1. Upload a course image
2. Delete the image object from file table (e.g. mongocli)
3. It's not possible to render list view for student/teacher/admin

```
TypeError: Cannot read property 'toObject' of null
    at model.<anonymous> (/home/daniel/Code/geli/api/build/src/models/Course.js:284:82)
    at Generator.next (<anonymous>)
    at fulfilled (/home/daniel/Code/geli/api/build/src/models/Course.js:4:58)
```
https://github.com/geli-lms/geli/blob/c016a39821be11c2a40b585275efa329235a096f/api/src/models/Course.ts#L319

I'm not sure if its possible to force this error somehow (i did but can't remeber how). Code above checks if course.image is not null but not if there is a matching object in file table (not null) before calling toObject.

## Acceptance criteria:
- [ ] No unsafe access

danielkesselberg:
Its possible that an course image does not exist as file anymore.
This pull requests wrap calls to picture objects with try catch.

## Description:

Closes #1053 

## Improvements
- Secure course picture upload/remove

### DIFF
--- a/api/src/controllers/CourseController.ts
+++ b/api/src/controllers/CourseController.ts
@@ -23,6 +23,7 @@ import ResponsiveImageService from '../services/ResponsiveImageService';
 import {IResponsiveImageData} from '../../../shared/models/IResponsiveImageData';
 
 import { Picture } from '../models/mediaManager/File';
+import {extractSingleMongoId} from '../utilities/ExtractMongoId';
 
 const coursePictureUploadOptions = {
   storage: multer.diskStorage({
@@ -713,10 +714,9 @@ export class CourseController {
       throw new NotFoundError();
     }
 
-    try {
-      (await Picture.findOne(course.image)).remove();
-    } catch (err) {
-      // go ahead if we cannot delete the course picture
+    const picture = await Picture.findById(extractSingleMongoId(course.image));
+    if (picture) {
+      picture.remove();
     }
 
     await Course.updateOne({_id: id}, {$unset: {image: 1}});
@@ -783,10 +783,9 @@ export class CourseController {
       throw new BadRequestError('Forbidden format of uploaded picture: ' + mimeFamily);
     }
 
-    try {
-      (await Picture.findById(course.image)).remove();
-    } catch (err) {
-      // go ahead if we cannot delete the course picture
+    const picture = await Picture.findById(extractSingleMongoId(course.image));
+    if (picture) {
+      picture.remove();
     }
 
     const responsiveImageData = <IResponsiveImageData>JSON.parse(responsiveImageDataRaw.imageData);
--- a/api/src/controllers/CourseController.ts
+++ b/api/src/controllers/CourseController.ts
@@ -704,7 +704,7 @@ export class CourseController {
     @Param('id') id: string,
     @CurrentUser() currentUser: IUser) {
 
-    const course = await Course.findById(id).orFail(new NotFoundError()) as ICourseModel;
+    const course = await Course.findById(id).orFail(new NotFoundError());
 
     if (!course.checkPrivileges(currentUser).userCanEditCourse) {
       throw new ForbiddenError();
@@ -770,7 +770,7 @@ export class CourseController {
       @CurrentUser() currentUser: IUser) {
 
     // Remove the old picture if the course already has one.
-    const course = await Course.findById(id).orFail(new NotFoundError()) as ICourseModel;
+    const course = await Course.findById(id).orFail(new NotFoundError());
 
     if (!course.checkPrivileges(currentUser).userCanEditCourse) {
       throw new ForbiddenError();
--- a/api/src/controllers/CourseController.ts
+++ b/api/src/controllers/CourseController.ts
@@ -21,9 +21,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import ResponsiveImageService from '../services/ResponsiveImageService';
 import {IResponsiveImageData} from '../../../shared/models/IResponsiveImageData';
-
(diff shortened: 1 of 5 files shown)


api/src/controllers/CourseController.ts
```diff
@@ -702,13 +713,54 @@ export class CourseController {
       throw new NotFoundError();
     }
 
-    const picture = await Picture.findOne(course.image);
-    await picture.remove();
+    try {
+      (await Picture.findOne(course.image)).remove();
```
torss:
```ts
await Picture.findByIdAndRemove(course.image);
```
Wouldn't it be better / "more correct" to write it like this?

api/src/controllers/CourseController.ts
```diff
@@ -718,27 +770,26 @@ export class CourseController {
...
       await fs.unlinkSync(file.path);
-
       throw new BadRequestError('Forbidden format of uploaded picture: ' + mimeFamily);
     }
 
-    if (course.image) {
-      const picture = await Picture.findById(course.image);
-      await picture.remove();
+    try {
+      (await Picture.findById(course.image)).remove();
```
torss:
```ts
await Picture.findByIdAndRemove(course.image);
```
Wouldn't it be better / "more correct" to write it like this?

api/src/models/Course.ts
```diff
@@ -316,7 +316,13 @@ courseSchema.methods.forDashboard = async function (user: IUser): Promise<ICours
     name, active, description, enrollType
   } = this;
 
-  const image = (this.image) ? (await Picture.findById(this.image)).toObject() : null;
+  let image;
+
+  try {
+    image = (await Picture.findById(this.image)).toObject();
```
torss:
```ts
const picture = await Picture.findById(this.image);
const image = picture && picture.toObject();
```
Wouldn't it be better / "more correct" to write it like this or something similar?
_(Because right now the `try`/`catch` would also absorb any potential actual errors, other than just a missing `this.image`.)_

torss: Also I guess the `ICourseDashboard` should then probably make `image: IPicture;` optional (`image?: IPicture;`).

api/src/controllers/CourseController.ts
danielkesselberg: Yes but the `remove` middleware is not executed when we use `findByIdAndDelete`. https://github.com/geli-lms/geli/blob/develop/api/src/models/mediaManager/File.ts#L46

We could move the logic into a dedicated method and register to `findOneAndDelete` middleware as well.

api/src/models/Course.ts
danielkesselberg: I need to catch various cases here. `this.image` is null and `Picture.findById(this.image)` is null. Looked more complicated to me to write this as if-else. Our goal should be that "something is wrong with the course image" does not stop the course view from working.

torss: Maybe I'm mistaken, but I thought `await Picture.findById(this.image)` should return `undefined` when `this.image` is `null` (or `undefined`)?

api/src/controllers/CourseController.ts
torss: Ah, alright! Then maybe just
```ts
const picture = await Picture.findById(this.image);
if (picture) {
  picture.remove();
}
```
would be sensible _(unless I'm mistaken re. `await Picture.findById(null)`, see below)_.

api/src/models/Course.ts
danielkesselberg: Yes and it fails on `toObject` then :thinking: (see the issue https://github.com/geli-lms/geli/issues/1053)

torss: Not with the aforementioned
```ts
const image = picture && picture.toObject();
```
part I would hope! ;P
If `await Picture.findById(this.image)` returns `undefined` then this way `image` will also be `undefined`, solving the problem without `try`/`catch`. Of course, if you don't like the code style with `&&`, we could just use an `if`-statement or whatever. To clarify, my only concern with your current solution is the `try`/`catch`, since it could theoretically also silence real errors in the future, that's all.

danielkesselberg: `const image = picture && picture.toObject as any;` would work. `ICourseDashboard` is a bit wrong about image. toObject is not castable to `IPicture` so we should use another typehint there.

torss: Hmm `interface IPictureModel extends IPicture, IFileModel`, with `interface IPicture extends IFile`, but the `pictureSchema` only defines the two direct properties of `IPicture` (`breakpoints` & `mimeType`), maybe that's (part of) the problem.

torss: Anyway, if you think that should have its own issue the `as any` solution _(or whatever replaces the `try`/`catch`)_ is fine by me.

api/src/controllers/CourseController.ts
```diff
@@ -718,27 +770,25 @@ export class CourseController {
       @CurrentUser() currentUser: IUser) {
 
     // Remove the old picture if the course already has one.
-    const course = await Course.findById(id);
+    const course = await Course.findById(id).orFail(new NotFoundError()) as ICourseModel;
```
danielkesselberg: @torss could you have a look at this? I dont like the typecast here but webstorm complains about `IUserModel` :confused: 

image

torss: Hm, don't see that issue here with `Visual Studio Code`. Any idea why that's happening?

Issue #922, opened 2018-10-28
Translations for Snackbars etc


MichaelNarkus:
# User Story

### I want:
snackbars and other messages to also be translated to my selected language

### so that:
the displayed language is consistent 

## Acceptance criteria:
Check (and Fix) Typescript files for:
- [ ] About
- [ ]  Admin
- [ ]  Authenfication
- [ ] Course
- [ ] File
- [ ] Lecture
- [ ] Messaging
- [ ] Notification
- [ ] Privacy
- [ ] Progress
- [ ] Report
- [ ] Shared
- [ ] Start
- [ ] Unit
- [ ] User

MichaelNarkus:
## TranslateService
As discussed in yesterdays conference it is indeed not necessary to define the default language in every file.  It is though necessary to import the TranslateService within the typescript-files when snackbars or other runtime-generated messages need to be translated, which imho should be fine. 
Similar to the existing or within SharedModule the translateService could be imported through a Module. This would limit the import to files that are fe. not already including SharedModule.  However this does not solve the original problem we had with the translations in typescript files : subscribe

## Subscribe vs await
According to the ngx-translate-documentation this is the way it should be done:

translate.get('teststring').subscribe((res: string) => {
    console.log(res);
    //=> 'hello world'
});
Using await though is another way to receive translations from the translateService but reserved for async functions only.

Efforts to translate all needed strings in fe. ngOnInit resulted in sometimes missing strings in components with multiple snackbars or status updates.

## Creating new loader
Creating a new loader and implementing the existing TranslateLoader unfortunately doesn't seem to solve our problem hence it also has to return an Observable. 
'The only required method is getTranslation that must return an Observable.' (ngx-translate-docs)

PR #1074, opened 2018-12-13
Add service for translatable snackbars

danielkesselberg:
## Description:

Ref #922 (not close because there is many code to migrate)

## Improvements
- Add a service that translates a message and shows a snackbar then
- Tests for SnackBarService are working again


danielkesselberg:
As discussed on monday: I think multiple keys for snackbars are not required. Merged the special case into three keys.

PR #1186, opened 2019-03-20
Bugfix/#1116 ProgressController vulnerabilities

### LINKED ISSUE #1116: 🐛 BUG: ProgressController vulnerabilities
There are some problems in the `ProgressController`, including vulnerabilities:

* The `GET` routes `{get} /api/progress/units/:id` (`getUnitProgress`) & `{get} /api/progress/courses/:id` (`getCourseProgress`) & `{get} /api/progress/users/:id` (`getUserProgress`) don't check that the `@CurrentUser` is authorized to access the route.
* These `GET` routes also could be refactored to use `async`/`await`.
* `{post} /api/progress/` (`createProgress`) & `{put} /api/progress/:id` (`updateProgress`) also don't check for `@CurrentUser` authorization.
* Shared route functionality can be factored out.
* Error messages don't utilize the central `errorCodes` file.
* In `{post} /api/progress/` the `currentUser?: IUser` parameter is marked as optional, but might actually not be optional.

[**Addendum:** Further flaws / opportunities were discovered while working on this issue, they are listed in via the following `Acceptance criteria`.]

## Acceptance criteria:

- [x] Write an access success unit test for `{get} /api/progress/units/:id`.
- [x] Write an access success unit test for `{get} /api/progress/courses/:id`.
- [x] Write an access success unit test for `{get} /api/progress/users/:id`.
- [x] ~~Check whether `{get} /api/progress/units/:id` is actually used anywhere in the front-end.~~ `getUnitProgress` from `ProgressService` is used in the `code-kata-unit` & `task-unit` component.
- [x] ~~Check whether `{get} /api/progress/courses/:id` is actually used anywhere in the front-end.~~ `getCourseProgress` from `ProgressService` is apparently unused.
- [x] ~~Check whether `{get} /api/progress/users/:id` is actually used anywhere in the front-end.~~ `getUserProgress` from `ProgressService` is apparently unused.
- [x] Possibly ~~remove or~~ disable `{get} /api/progress/courses/:id` since it appears to be unused.
- [x] Possibly ~~remove or~~ disable `{get} /api/progress/users/:id` since it appears to be unused.
- [x] Write a fitting access denial unit test for `{get} /api/progress/units/:id`.
- [x] Write a `{get} /api/progress/units/:id` unit test to check that `student` requests only respond with their own data.
- [x]  **[Done, but obviated]** ~~Write a fitting access denial unit test for `{get} /api/progress/courses/:id`.~~
- [x] **[Obviated]** ~~Write a fitting access denial unit test for `{get} /api/progress/users/:id`.~~
- [x] Write a fitting access denial unit test for `{post} /api/progress/`.
- [x] Write a fitting access denial unit test for `{put} /api/progress/:id`.
- [x] Secure the `{get} /api/progress/units/:id` route by checking the `@CurrentUser` authorization.
- [x] Secure the `{get} /api/progress/units/:id` route by filtering `student` requests to only respond with their own data.
- [x] **[Obviated]** ~~Secure the `{get} /api/progress/courses/:id` route by checking the `@CurrentUser` authorization.~~
- [x] **[Obviated]** ~~Secure the `{get} /api/progress/users/:id` route by checking the `@CurrentUser` authorization.~~
- [x] Secure the `{post} /api/progress/` route by checking the `@CurrentUser` authorization (course access).
- [x] Secure the `{put} /api/progress/:id` route by checking the `@CurrentUser` authorization (course access & `progress.user` id).
- [x] Refactor the `GET` routes to use `async`/`await`.
- [x] **[Obviated]** ~~Factor out shared route functionality where possible.~~
- [x] Utilize the central `errorCodes` file for the error message~~s~~.
- [x] ~~Check /~~ Fix the optionality of the `{post} /api/progress/` `currentUser?: IUser` parameter.
- [x] Maybe the `{post} /api/progress/` can be obviated by a more flexible (and idempotent) iteration of `{put} /api/progress/:id`? (Yes it apparently can, done.)
- [x] Refactor (the existing) unit tests.
- [x] `ProgressService.getUnitProgress` only uses the first item from the list returned by the `{get} /api/progress/units/:id` route and there is no reason why that route should return a list anyway, so it should return a single progress object instead.
- [x] **[Opportunistic]** Fix `TaskUnitComponent.validate` `validationMode` reset.
- [x] **[Opportunistic]** Fix `CodeKataComponent` `progress.code` loading.

## Additional info:
The security problems were discovered during the recent #853 audits, [see `ProgressController`](https://github.com/geli-lms/geli/issues/853#issuecomment-435216631).

torss:
## Description:

Closes #1116

Fixes the `ProgressController` vulnerabilities and conducts some related opportunistic improvements.

## Improvements
- Closes the `ProgressController` vulnerabilities, by checking the `@CurrentUser` via `course.checkPrivileges`.
- Adds `ProgressController` `GET` unit tests & access denial tests in general.
- Also refactors the (existing) unit tests.
- Extends the `ProgressController` `PUT` route to handle both creation and updates, obviating (and thus removing) the `POST` route entirely.
- ~~Deactivated~~ Removes the `ProgressController` `GET` routes for `/courses/` & `/users/` since they were apparently unused.
- The (remaining) `GET` route now only returns a single (possibly empty) progress object instead of a list thereof, because that didn't make sense and the front-end only utilized the first list element anyway.
- Fixes the `TaskUnitComponent.validate` `validationMode` reset.
- Fixes the `CodeKataComponent` `progress.code` loading.
- Other minor fixes / refactoring (e.g. `errorCodes` usage).

_(For more details see the issue's `Acceptance criteria` list or, of course, commit messages.)_

## Known Issues:
_NONE_


api/src/controllers/ProgressController.ts
```diff
@@ -25,10 +26,10 @@ export class ProgressController {
    *
    * @apiParam {String} id Unit ID.
    *
-   * @apiSuccess {Progress[]} progresses Progress.
+   * @apiSuccess {Progress} progress Progress data or an empty object if no data is available.
```
kesselb: https://github.com/geli-lms/geli/blob/656221c1adcdd891f14158bfe002698285651b09/app/webFrontend/src/app/shared/services/data/progress.service.ts#L11-L14 Is angular expecting an array or object?

api/src/controllers/ProgressController.ts
```diff
@@ -100,11 +108,14 @@ export class ProgressController {
    *         "id": "5ab2b9516fab4a3ae0cd6737"
    *     }]
    */
+  // This route has been disabled since it appears to be unused and insufficiently secured.
```
kesselb: Remove it :sunglasses: 

https://github.com/geli-lms/geli/blob/656221c1adcdd891f14158bfe002698285651b09/app/webFrontend/src/app/shared/services/data/progress.service.ts#L16-L22

Also this one

api/src/controllers/ProgressController.ts
```diff
@@ -144,71 +155,17 @@ export class ProgressController {
...
```
kesselb: Remove it :sunglasses: 

https://github.com/geli-lms/geli/blob/656221c1adcdd891f14158bfe002698285651b09/app/webFrontend/src/app/shared/services/data/progress.service.ts#L24-L30

Also this one

api/src/controllers/ProgressController.ts
torss: Yes, but `getUnitProgress` has already been changed accordingly:
https://github.com/geli-lms/geli/blob/b59ff2e7a1c45a5a2a63911d115cfa94c6e0abef/app/webFrontend/src/app/shared/services/data/progress.service.ts#L11-L15

api/src/controllers/ProgressController.ts
torss: Ok, I will remove it completely!
_(BTW `getCourseProgress` was already commented out as well.)_

api/src/controllers/ProgressController.ts
torss: Ok, I will remove it completely!
_(BTW `getUserProgress` was already commented out as well.)_

torss: Done.

api/src/controllers/ProgressController.ts
torss: Done.

api/src/controllers/ProgressController.ts
kesselb: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined i guess we should return the empty object or undefined then. Could we use a 404 exception if there is no progress? That would be easier to handle in progress.service

torss: > https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined 

That's hilarious, look at the headers above: https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#stop-reading-immediately
`[...] These are Coding Guidelines for Contributors to TypeScript [...] AGAIN: This is NOT a prescriptive guideline for the TypeScript community [...]` xD

Anyway, yeah, in my personal JavaScript stuff I tend to always use `undefined` and I just left it as `null` here because that's how it was written in the original code.

Anyway², to get to the actual issue: Originally I returned `null` in the `GET` route if no data can be found. This means that the response will have the status code `204` (No Content), which makes sense, but I changed it to the way it is now since the unit tests would have to be adapted otherwise, and more importantly because I wasn't sure whether this would cause problems in the front-end.
But this (`204` response) seems like a more sensible solution, so I will try that again now...

torss: ...or not. I have searched a bit and `204` isn't really meant for this use case, neither is `404`. E.g. https://medium.com/@santhoshkumarkrishna/http-get-rest-api-no-content-404-vs-204-vs-200-6dd869e3af1d advocates `200` with an empty body instead, so the way it is right now should actually be fine from that perspective. Merge?

app/webFrontend/src/app/shared/services/data/progress.service.ts
```diff
@@ -9,23 +9,14 @@ export class ProgressService extends DataService {
   }
 
   async getUnitProgress<T>(unitId: string): Promise<T> {
```
kesselb: Could you check the return type here? We are not returning a promise here.

torss: That should be correct, `async` implicitly wraps the return value in a `Promise`. See for example this random search result: https://basarat.gitbooks.io/typescript/docs/async-await.html ("async function always returns a Promise")

PR #1189, opened 2019-03-20
Remove migrate-mongoose dependency

kesselb:
## Description:

`migrate-mongoose` is not used anymore. 

## Improvements
- No more security warnings from github

## Known Issues:
_NONE_


kesselb:
> Can you verify this?

Yes. Looks like 'bson' is missing. Not sure if we need bson at all. This objectId class is also part of mongoose.

PR #1191, opened 2019-03-21
Bugfix/#1190 UnitController vulnerabilities

### LINKED ISSUE #1190: 🐛 BUG: UnitController vulnerabilities
The `UnitController` routes have some vulnerabilities:

* `{get} /api/units/:id`: Doesn't check for `@CurrentUser` authorization. _(Likely not critical since the unit ID must be known.)_
* The other routes, `{post} /api/units/` & `{put} /api/units/:id` & `{delete} /api/units/:id`: Don't check for `@CurrentUser` authorization beyond `@Authorized(['teacher', 'admin'])`, so a `teacher`-level attacker could maliciously add / update / delete units to / of any course.

Related opportunistic improvements could also be conducted.

## Acceptance criteria:

- [x] **[Opportunistic]** Write a success unit test for `{get} /api/units/:id`.
- [x] **[Opportunistic]** Write a success unit test for `{delete} /api/units/:id`.
- [x] Write an access denial unit test for `{get} /api/units/:id`.
- [x] Write an access denial unit test for `{post} /api/units/`.
- [x] Write an access denial unit test for `{put} /api/units/:id`.
- [x] Write an access denial unit test for `{delete} /api/units/:id`.
- [x] Secure `{get} /api/units/:id` via `course.checkPrivileges`.
- [x] Secure `{post} /api/units/` via `course.checkPrivileges`.
- [x] Secure `{put} /api/units/:id` via `course.checkPrivileges`.
- [x] **[Opportunistic]** Refactor `{delete} /api/units/:id` to use `async`/`await`.
- [x] ~~**[Opportunistic]** Fix `{delete} /api/units/:id`'s `DeprecationWarning: collection.remove is deprecated. Use deleteOne, deleteMany, or bulkWrite instead.`.~~ Nevermind, just leave it be for now due to the missing middleware support: https://github.com/Automattic/mongoose/issues/7195
- [x] Secure `{delete} /api/units/:id` via `course.checkPrivileges`.
- [x] Factor out shared code using `course.checkPrivileges`.
- [x] ~~**[Opportunistic]** Return empty object on success in `{post} /api/units/` if possible/sensible.~~
- [x] ~~**[Opportunistic]** Return empty object on success in `{put} /api/units/:id` if possible/sensible.~~
- [x] **[Opportunistic]** Return empty object on success in `{delete} /api/units/:id` if possible/sensible.
- [x] **[Opportunistic]** Utilize `errorCodes`.
- [x] **[Opportunistic]** Refactor unit test code. (Done, but could be refactored further.)

## Additional info:

Apparently the `UnitController` doesn't have a directly dedicated unit test file, but the `codeKataUnit.ts` does contain tests for it, so the new tests could just be introduced there.

The security problems were discovered during the recent #853 audits, [see `UnitController`](https://github.com/geli-lms/geli/issues/853#issuecomment-435950164).

torss:
## Description:

Closes #1190

Fixes `UnitController` vulnerabilities.

## Improvements:
- Closed vulnerabilities for all 4 routes via `course.checkPrivileges`.
- Added `GET` & `DELETE` route unit tests for status code `200`.
- Added status code `403` (not authorized to view / edit course) unit tests for all 4 routes.
- Some related unit test refactoring _(could definitely be refactored further)_.
- `DELETE` route now returns a completely empty object _(very minor change)_.
- Other minor refactoring, such as `async`/`await` usage and `errorCodes` refactoring.

## Known Issues:
The `DELETE` route uses `await unit.remove()`, which gives a `DeprecationWarning: collection.remove is deprecated. Use deleteOne, deleteMany, or bulkWrite instead.`, but `deleteOne` is still missing middleware support (which `Unit` makes use of): Automattic/mongoose#7195 & https://mongoosejs.com/docs/api.html#model_Model.deleteOne
_This problem wasn't introduced by this PR, I just noticed it while running the unit tests, but left it be due to the middleware regression._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `UnitController` status code `403` (not authorized to view / edit course) unit tests for all routes. [#1190](https://github.com/geli-lms/geli/issues/1190)
 
 ### Changed
-- Extended `ProgressController` `PUT` route to handle both creation and updates. [#1116](https://github.com/geli-lms/geli/issues/1116)
-- Refactored `ProgressController` unit tests in general. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- Extend `ProgressController` `PUT` route to handle both creation and updates. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- Refactor `ProgressController` unit tests in general. [#1116](https://github.com/geli-lms/geli/issues/1116)
 - Instead of a list of progress data, the `ProgressController` `GET` route now responds with a single progress object or an empty object if no data can be found. [#1116](https://github.com/geli-lms/geli/issues/1116)
 
 ### Removed
@@ -31,8 +31,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `CodeKataComponent` `progress.code` loading. [#1116](https://github.com/geli-lms/geli/issues/1116)
 
 ### Security
-- Closed `ProgressController` vulnerabilities. [#1116](https://github.com/geli-lms/geli/issues/1116)
-- Closed `UnitController` vulnerabilities. [#1190](https://github.com/geli-lms/geli/issues/1190)
+- Closes `ProgressController` vulnerabilities. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- Closes `UnitController` vulnerabilities. [#1190](https://github.com/geli-lms/geli/issues/1190)
 
 ## [[0.8.4](https://github.com/geli-lms/geli/releases/tag/v0.8.4)] - 2018-12-20 - WS 18/19 ❄️-Release
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `CodeKataComponent` `progress.code` loading. [#1116](https://github.com/geli-lms/geli/issues/1116)
 
 ### Security
-- Closes `ProgressController` vulnerabilities. [#1116](https://github.com/geli-lms/geli/issues/1116)
-- Closes `UnitController` vulnerabilities. [#1190](https://github.com/geli-lms/geli/issues/1190)
+- Close `ProgressController` vulnerabilities. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- Close `UnitController` vulnerabilities. [#1190](https://github.com/geli-lms/geli/issues/1190)
 
 ## [[0.8.4](https://github.com/geli-lms/geli/releases/tag/v0.8.4)] - 2018-12-20 - WS 18/19 ❄️-Release
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Translatable SnackBarService. [#922](https://github.com/geli-lms/geli/issues/922)
 - `ProgressController` `GET` unit tests & access denial tests in general. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- `UnitController` `GET` & `DELETE` route unit tests for status code `200`. [#1190](https://github.com/geli-lms/geli/issues/1190)
+- `UnitController` status code `403` (not authorized to view / edit course) unit tests for all routes. [#1190](https://github.com/geli-lms/geli/issues/1190)
 
 ### Changed
 - Extended `ProgressController` `PUT` route to handle both creation and updates. [#1116](https://github.com/geli-lms/geli/issues/1116)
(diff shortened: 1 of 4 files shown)


api/src/controllers/UnitController.ts
```diff
@@ -12,6 +13,20 @@ import {IUser} from '../../../shared/models/IUser';
 @UseBefore(passportJwtMiddleware)
 export class UnitController {
 
+  static async getUnitFor (unitId: string, currentUser: IUser, privilege: 'userCanViewCourse' | 'userCanEditCourse') {
+    const unit = await Unit.findById(unitId);
+    if (!unit) {
```
kesselb: orFail() should work here

api/src/controllers/UnitController.ts
```diff
@@ -12,6 +13,20 @@ import {IUser} from '../../../shared/models/IUser';
 @UseBefore(passportJwtMiddleware)
 export class UnitController {
 
+  static async getUnitFor (unitId: string, currentUser: IUser, privilege: 'userCanViewCourse' | 'userCanEditCourse') {
```
kesselb: Why static? This method is only used here

api/src/controllers/UnitController.ts
torss: Thanks, I forgot about that being an option, will change it!

api/src/controllers/UnitController.ts
torss: It's static because it doesn't need access to `this`. But yeah, it probably should also be `protected`, will add that...

torss: Done!

api/src/controllers/UnitController.ts
torss: [Done](https://github.com/geli-lms/geli/pull/1191/commits/745263724cc3e299325d12695ece70882b10f485)!

api/src/controllers/UnitController.ts
kesselb:
```suggestion
  private async getUnitFor (unitId: string, currentUser: IUser, privilege: 'userCanViewCourse' | 'userCanEditCourse') {
```

torss: It currently is `protected static` and I don't see how making it not-`static` would be better. _(And `protected` vs `private` is pretty much a moot point; in fact if a hypothetical deriving class would exist, the `protected` would make more sense here.)_

kesselb: https://stackoverflow.com/questions/2671496/java-when-to-use-static-methods
https://stackoverflow.com/questions/24289070/why-we-should-not-use-protected-static-in-java

`private` is the right pick here. If you use this from another Controller there is something wrong with the architecture.

torss: Hm I disagree with applying the reasoning seen in some of these Java stackoverflow answers here; e.g. note that `protected` in Java isn't quite the same as `protected` in TypeScript. My reasoning for `static` is that the function doesn't use any instance members or `this`.

But your `If you use this from another Controller there is something wrong with the architecture.` argument makes sense and I don't care too much about the `static` stuff, so alright, I will change all of the methods to non-`static` & `private`. ;)

torss: Done!

api/src/controllers/UnitController.ts
```diff
@@ -158,32 +163,39 @@ export class UnitController {
...
+    const unit = await this.getUnitFor(id, currentUser, 'userCanEditCourse');
+
+    await Lecture.updateMany({}, {$pull: {units: id}});
+    await unit.remove();
+    return {};
+  }
+
+  private async getUnitFor (unitId: string, currentUser: IUser, privilege: 'userCanViewCourse' | 'userCanEditCourse') {
+    const unit = await Unit.findById(unitId);
+    if (!unit) {
```
kesselb: `orFail` :thinking:

torss: Argh, thanks, good catch - must have copied the old version! :)

torss: [Fixed](https://github.com/geli-lms/geli/pull/1191/commits/b55a7726c96a77364e5a9c02bc84ac0c2fea30b4)!

CHANGELOG.md
```diff
@@ -30,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Security
 - Closed `ProgressController` vulnerabilities. [#1116](https://github.com/geli-lms/geli/issues/1116)
+- Closed `UnitController` vulnerabilities. [#1190](https://github.com/geli-lms/geli/issues/1190)
```
kesselb: We used the present for changelog entries (compare 0.8.4)

torss: Right, will change it...

torss: [Done](https://github.com/geli-lms/geli/pull/1191/commits/96137cc53d7091480e883761b60ef5f2c3e6bbd7)!

torss: [Fixed](https://github.com/geli-lms/geli/pull/1191/commits/ea91a7c8d965ec566c2fffbdeca0afa674ab9e87)!

PR #1195, opened 2019-03-22
Remove file cache from DownloadController

### LINKED ISSUE #913: PDF export fails randomly

kesselb:
## Description:

Closes #913 

If a user exports a course we cache the result for about an hour. This is a problem because there is no way to invalidate the cache. We don't reuse the exported result after this pr.

## Improvements
- `userCanExportCourse` used by single and individual export
- Export pdf as buffer and append to archive directly (a temporary file is no longer required)
- Drop `node-file-cache` dependency
- Add type definitions for `archiver` and `html-pdf` 

## Known Issues:
- Code could be improved


PatrickSkowronek:
@kesselb Code looks good. Last year it was the best way we found, but this is clearly better. Note this needs to be done in assignment controller as well.

api/test/integration/download.ts
```diff
@@ -188,6 +192,31 @@ describe('DownloadFile', () => {
...
+    //   const teacher = await User.findById(course.teachers.pop());
+    //   const testData = await prepareTestData(course);
+    //   const res = await chai.request(app)
+    //     .post(BASE_URL + '/pdf/individual')
+    //     .set('Cookie', `token=${JwtUtils.generateToken(teacher)}`)
+    //     .send(testData)
+    //     .catch(err => err.response);
+    //   expect(res).to.have.status(200);
+    //   expect(res).to.be.json;
+    // }).timeout(30000);
```
dboschm: Is there a special reason for this out commented code?

api/test/integration/download.ts
```diff
@@ -188,6 +192,31 @@ describe('DownloadFile', () => {
       expect(res).to.have.status(200);
       expect(res).to.be.json;
     }).timeout(30000);
+    it('should pass (student)', async () => {
+      const course = await FixtureUtils.getRandomCourseWithAllUnitTypes();
+      const student = await User.findById(course.students.pop());
```
torss: This might just be a nitpick, but it seems like `course.students.pop()` is logically incorrect since `.pop()` removes the returned element from the (`course.students`) array, so maybe change it to e.g. `course.students[0]`? _(Same below in the commented-out code for `course.teachers`.)_

api/test/integration/download.ts
torss: Weird, looks to me that the `// For some reason there is no teacher for a course.` shouldn't happen, but it [evidently](https://travis-ci.com/geli-lms/geli/jobs/187906674) does. One could try `FixtureUtils.getRandomTeacherForCourse` instead of the `await User.findById(course.teachers[0])`, that might work because it takes the `courseAdmin` into account, but there already is a separate test for that - and it doesn't explain why `await User.findById(course.teachers[0])` doesn't work anyway.

torss: Ok so after multiple local test runs it seems that the issue is non-deterministic, i.e. it sometimes works and sometimes there is no teacher in the `teachers` array.

torss: Ohhh of course! The `FixtureLoader` simply assigns `course.teachers = await FixtureUtils.getRandomTeachers(0, 2, hash);`, so either 0, 1 or 2 teachers per course, I totally missed that earlier.

torss: Alright, so I think the test should be removed / combined with the `(course admin)` test via `FixtureUtils.getRandomTeacherForCourse`, I will commit something...

PR #819, opened 2018-08-01
WIP: Feature/#727 add guided dialog for adding a whitelist

lukas-schardt:
## Description:
Adds a guided dialog for creating a whitelist

Closes #509, closes #727 

## Improvements
- 

## Known Issues:
_NONE_

PR #1352, opened 2019-09-03
Remove fast-csv dependency

kesselb:
fast-csv was used to parse a csv with students from obs. Since https://github.com/geli-lms/geli/pull/819 the csv is parsed by the frontend.


kesselb:
@dependabot merge

